Issue #8, opened 2018-02-12
I think all the tests for fivethirtyeight need to be updated


oxinabox:
fivethirtyeight  updated their LICENSE.md.
Which is great.
But it breaks all the tests, since they are reference tests on exact output.

They should just need rerunning and told to overwrite.

PR #9, opened 2018-02-26
fivethirtyeight tests updated

SebastinSanty:
The tests are working perfectly now. I made some additional changes, which I am not sure if I am correct. I am stating my reasons for the same:

1) I changed the cdn from `rawgit.com` to `raw.githack.com` as the tests (and myself) were not able to access the urls returned by `generate(Github(), <url>)`
2) Though I completely skeptical about this, I removed the `get_docfile(GitHub(), mainpage, "LICENSE")` as the tests were failing because of this. Naturally so, there weren't any LICENSE file present at the location specified.


SebastinSanty:
Fixes  #8

oxinabox:
Nice work. It was apparently harder than I thought.
Seems like fivethirtyeight has made some changed to their repo files, and now finding the dataset name works on more of them, which is a nice side effect.
No doubt some of those changed also relate to why things were more difficult than expected.
Good job

re: 1, swapping rawgit to githack is fine. The readme also need to be updated though.

That repo does have a LISCENSE file though,
it is at https://github.com/fivethirtyeight/data/blob/master/LICENSE
Might need to do some further debugging to workout why that URL is not being found (and/or is not being correctly converted to a githack url).

SebastinSanty:
I reverted back to rawgit cdn. I think the rawgit server was down yesterday, which is why I was not able to access them. With these changes, everything comes back to normal and this PR is a easy fix as you had said earlier.

oxinabox:
Nice work.

oxinabox:
@SebastinSanty  have you joined the julia slack?
http://slackinvite.julialang.org/
In particular the #jsoc channel?

SebastinSanty:
Thanks for the leads. I have joined julia slack.

PR #12, opened 2018-03-01
Make DataDryad work

SebastinSanty:
This is a very crude way of getting the desired output.

### DIFF
--- a/src/DataDryad.jl
+++ b/src/DataDryad.jl
@@ -1,0 +1,42 @@
+immutable DataDryad <: DataRepo
+end
+
+base_url(::DataDryad) = "https://datadryad.org/mn/"
+
+function description(::DataDryad, mainpage)
+    desc = replace(text_only(first(matchall(sel".article-abstract", mainpage.root))), "Abstract ", "")
+    author = text_only(first(matchall(sel".pub-authors a", mainpage.root)))
+    license = getattr(first(matchall(sel".single-image-link", mainpage.root)), "href")
+    date = now()
+    paper = text_only(first(matchall(sel".citation-sample a", mainpage.root)))
+    dataset = replace(text_only(last(matchall(sel".publication-header p", mainpage.root))), "DOI: ", "")
+
+    final = escape_multiline_string("""
+    Author: $(author) et. al.
+    License: $(license)
+    Date: $(date)
+
+    $(desc)
+
+    Please cite this paper:
+    $(paper)
+    as well as this dataset:
+    $(dataset)
+    if you use this in your research.
+    """, "\$")
+end
+
+function get_urls(repo::DataDryad, page)
+    urls = []
+    paper = text_only(last(matchall(sel".publication-header p", page.root)))
+    download_url = replace(paper, "DOI: https://", "https://datadryad.org/mn/object/http://dx.") * "/1/bitstream"
+    push!(urls, download_url)
+    md5 = ("md5", text_only(getpage(replace(paper, "DOI: https://doi.org/", "https://datadryad.org/mn/checksum/doi:") * "/1").root))
+    push!(urls, md5)
+    urls
+end
+
+function data_fullname(::DataDryad, mainpage)
+    # mainpage = replace(mainpage, "resource", "mn/object")
+    text_only(first(matchall(sel".pub-title", mainpage.root)))
+end
--- a/src/DataDryad.jl
+++ b/src/DataDryad.jl
@@ -5,10 +5,16 @@ base_url(::DataDryad) = "https://datadryad.org/mn/"
 
 function description(::DataDryad, mainpage)
     desc = replace(text_only(first(matchall(sel".article-abstract", mainpage.root))), "Abstract ", "")
-    author = text_only(first(matchall(sel".pub-authors a", mainpage.root)))
+    author = ""
+    try
+        author = text_only(first(matchall(sel".pub-authors a", mainpage.root)))
+    catch
+        author = string(split(text_only(first(matchall(sel".pub-authors", mainpage.root))), ", ")[1])
+    end
     license = getattr(first(matchall(sel".single-image-link", mainpage.root)), "href")
-    date = now()
-    paper = text_only(first(matchall(sel".citation-sample a", mainpage.root)))
+    dateelem = matchall(sel".publication-header p", mainpage.root)
+    date = replace(text_only(dateelem[length(dateelem)-1]), "Date Published: ", "")
+    paper = text_only(first(matchall(sel".citation-sample", mainpage.root)))
     dataset = replace(text_only(last(matchall(sel".publication-header p", mainpage.root))), "DOI: ", "")
 
     final = escape_multiline_string("""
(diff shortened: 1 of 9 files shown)


oxinabox:
Nice work, I won't have time to review this in detail til tomorrow (or possibly Monday).
But a quick look says it looks pretty good.

oxinabox:
I've still not properly read this, but so you can start thinking about it.
And reading about abstract classes in julia (if you've not already).

the next step is to make

`abstract DataOne`
`struct DataDryad <: DataOne`

and have functionality for the other member nodes
https://www.dataone.org/current-member-nodes

by having another e.g. `struct OneShare <: DataOne` , `struct TERN <: DataOne`

src/DataDryad.jl
```diff
@@ -0,0 +1,42 @@
+immutable DataDryad <: DataRepo
+end
+
+base_url(::DataDryad) = "https://datadryad.org/mn/"
+
+function description(::DataDryad, mainpage)
+    desc = replace(text_only(first(matchall(sel".article-abstract", mainpage.root))), "Abstract ", "")
+    author = text_only(first(matchall(sel".pub-authors a", mainpage.root)))
+    license = getattr(first(matchall(sel".single-image-link", mainpage.root)), "href")
+    date = now()
```
oxinabox: This should be the date the data was created, not the current date.
It should be read from the `mainpage`

src/DataDryad.jl
```diff
@@ -0,0 +1,42 @@
+immutable DataDryad <: DataRepo
+end
+
+base_url(::DataDryad) = "https://datadryad.org/mn/"
+
+function description(::DataDryad, mainpage)
+    desc = replace(text_only(first(matchall(sel".article-abstract", mainpage.root))), "Abstract ", "")
+    author = text_only(first(matchall(sel".pub-authors a", mainpage.root)))
```
oxinabox: an Extracting out just the first author and then saying `$(author) et. al.` 
is a good first approximation of the correct thing to say,
but it is not quiet correct.

This isn't blocking as far as merging this goes.
But it we merge it like it is will need to open an issue to remind us to fix it later.
(which is fine)

The correct is to do something like as described in
http://blog.apastyle.org/apastyle/2011/11/the-proper-use-of-et-al-in-apa-style.html

src/DataDryad.jl
```diff
@@ -0,0 +1,42 @@
...
+    license = getattr(first(matchall(sel".single-image-link", mainpage.root)), "href")
+    date = now()
+    paper = text_only(first(matchall(sel".citation-sample a", mainpage.root)))
```
oxinabox: Is it possible/easy to get the name of the paper and it's author list?

Ideally we would say something like:

`von Wettberg EJB, Chang PL, Başdemir F, Carrasquila-Garcia N, Korbu L, Moenga SM, Bedada G, Greenlon A, Moriuchi KS, Suryawanshi V, Cordeiro MA, Noujdina NV, Dinegde KN, Abbas Shah Sani SG, Getahun T, Vance L, Bergmann E, Lindsay D, Mamo BE, Warschefsky EJ, Dacosta-Calheiros E, Marques E, Yilmaz MA, Cakmak AM, Rose J, Migneault A, Krieg CP, Saylak S, Temel H, Friesen ML, Silor E, Akhmetov Z, Ozelik H, Kholova J, Can C, Gaur P, Yildirim M, Sharma H, Vadez V, Tesfaye K, Woldemedhin AF, Tar'an B, Aydogan A, Bukun B, Penmetsa RV, Berger J, Kahraman A, Nuzhdin SV, Cook DR (2018) Ecology and genomics of an important crop wild relative as a prelude to agricultural innovation. Nature Communications 9: 649. https://doi.org/10.1038/s41467-018-02867-z`

And similar for the dataset.

If it is not easy to get that into I am fine to merge as is with regards to this.

src/DataDryad.jl
```diff
@@ -0,0 +1,42 @@
...
+    if you use this in your research.
+    """, "\$")
+end
+
+function get_urls(repo::DataDryad, page)
+    urls = []
+    paper = text_only(last(matchall(sel".publication-header p", page.root)))
+    download_url = replace(paper, "DOI: https://", "https://datadryad.org/mn/object/http://dx.") * "/1/bitstream"
+    push!(urls, download_url)
+    md5 = ("md5", text_only(getpage(replace(paper, "DOI: https://doi.org/", "https://datadryad.org/mn/checksum/doi:") * "/1").root))
```
oxinabox: This isn't going to workout right:
the download urls and the md5 s are a different argument.
This produces: (in the example below)

```
Any["https://datadryad.org/mn/object/http://dx.doi.org/10.5061/dryad.74699/1/bitstream", ("md5", "bc96dba38d8659f42527cc22ff7f6e3b")]
```

But what we want is:

```
Any["https://datadryad.org/mn/object/http://dx.doi.org/10.5061/dryad.74699/1/bitstream"],
Any[("md5", "bc96dba38d8659f42527cc22ff7f6e3b")]
```

Fixing this is going to mean doing a bit more reworking, than is needed in this PR.
we should separate `get_checksums` from `get_urls`,
and then modify the main method around
https://github.com/oxinabox/DataDepsGenerators.jl/blob/master/src/DataDepsGenerators.jl#L31
to call it.
But we'll also need to provide correct handling for the other generators, which do not (currently),
define  `get_checksums`.

Further to that, I'm smelling an a bit of a code-smell in my code in that area.
So leave it for now, by just commenting out the md5 stuff,
and I'll propose a refactoring PR after this is done.

src/DataDryad.jl
```diff
@@ -0,0 +1,42 @@
...
+    as well as this dataset:
+    $(dataset)
+    if you use this in your research.
+    """, "\$")
+end
+
+function get_urls(repo::DataDryad, page)
+    urls = []
+    paper = text_only(last(matchall(sel".publication-header p", page.root)))
+    download_url = replace(paper, "DOI: https://", "https://datadryad.org/mn/object/http://dx.") * "/1/bitstream"
```
oxinabox: This isn't going to work for datasets with multiple files
e.g. https://datadryad.org/resource/doi:10.5061/dryad.f9s4424

Probably good to add that one as an extra text

oxinabox (changes requested):
Nice work, pretty good for your second (?) bit of julia coding.

I've left a bunch of comments.

The ones that need fixing before I can merge this are 

 - making it work for multiple files in on dataset,
eg. https://datadryad.org/resource/doi:10.5061/dryad.f9s4424
 - commenting out the md5 stuff (for now)
 - Fixing the date

src/DataDryad.jl
SebastinSanty: A separate logic will have to be written for this. Perhaps as you said, I'll file a separate issue for this.

src/DataDryad.jl
SebastinSanty: Implemented.

src/DataDryad.jl
SebastinSanty: Yes, I was skeptical of changing the structure of DataDepsGenerators, which I would have to do, if I was implementing in the format you suggested. That's why, I did it this way.

src/DataDryad.jl
SebastinSanty: Implemented.

src/DataDryad.jl
SebastinSanty: Implemented.

oxinabox:
Do you know why CI has started to fail?

oxinabox:
I am almost completely satisfied that this successfully
achieves the goal of being able to generate from DataDyrad.
But it does nothing towards the goal of generating from DataOne.
Still it is one thing done.
It will be easier to do  DataOne (including an alternate path to Dyrad) with this complete for comparison of outputs.


Could you add a test that checks that these generators work?

Something like adding `DataDeps` to the `/test/REQUIRE`.
and then making a new test file, which has something like:

```
using DataDeps
@testset "DataDyrad drought" begin
    @include("references/DataDryad Drought.txt") #register the data dep

    # Some sanity checks that we can use this generated DataDep

    intensities_csv = datadep"Data from Drought intensification drives turnover of structure and function in stream invertebrate communities/Drought intensities.csv"

    @test length(collect(eachline(intensities_csv))) > 0
end
```

For each of the 3 test cases.

We really should also have these for the other generated datadeps.

oxinabox (approved):

PR #13, opened 2018-04-30
Update references for fivethirtyeight and DataDryad

SebastinSanty:


oxinabox:
Thanks.
The downside of live integration tests / reference tests.

On the upside, they are trivial to write and easy to check.

PR #14, opened 2018-04-30
Update .travis.yml to allow failures on julia nightly

SebastinSanty:


oxinabox:
Oops, I must have missed when uncommenting lines.
Thanks

PR #15, opened 2018-05-07
Fix Github.jl dataset urls

SebastinSanty:
The `commit` variable doesn't work at the moment. Anyways `master` branch is a viable option.


oxinabox:
hmmm,
One actually wants to pin to a particular commit.
For 2 reasons

 1. We want to generate for This data, not for what ever is here later -- that could change. Most stable=more good
 2. Rawgit shouldn't actually support using a named branch with the http://cdn.rawgit.com.  See https://rawgit.com/ (on the left). And we want to use https://cdn.rawgit.com
however master is a moving target and is not really ideal.


So I think the action here is to findout *why* the current is not working and fix that.
My bad for not picking up the tests being wrong earlier

test/references/buzzfeed primates.txt
```diff
@@ -19,5 +19,5 @@ RegisterDataDep(
...
+    Any["https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/README.md", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/facility.json", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/facility.tsv", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/facility.txt", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/primates.r", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/report.json", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/report.tsv", "https://cdn.rawgit.com/BuzzFeedNews/2015-07-primates/tree-commit/report.txt"]
```
oxinabox: This didn't seem to be correct, I see no commit-sha in the URL

src/GitHub.jl
```diff
@@ -39,10 +39,15 @@ end
 
 
 function get_cdn_url_converter(mainpage)
-    commit = strip(text_only(matchall(sel".commit-tease-sha", mainpage.root)))
+    commit = match(r"\b[0-9a-f]{5,40}\b", matchall(sel".commit-tease", mainpage.root)[1].attributes["src"])
     function(urlsub)
         ret = "https://cdn.rawgit.com"*urlsub
-        replace(ret, "blob/master", commit, 1)
+        if commit === nothing
```
oxinabox: is this required? Does that regex ever fail?
If so is falling back to "master", valid -- does it actually create an URL that resolves.
(I think it does not)

src/GitHub.jl
```diff
@@ -39,10 +39,15 @@ end
 
 
 function get_cdn_url_converter(mainpage)
-    commit = strip(text_only(matchall(sel".commit-tease-sha", mainpage.root)))
+    commit = match(r"\b[0-9a-f]{5,40}\b", matchall(sel".commit-tease", mainpage.root)[1].attributes["src"])
```
oxinabox: break this into two lines:
```
commit_ele =  matchall(sel".commit-tease", mainpage.root)[1]
commit = match(r"\b[0-9a-f]{5,40}\b", commit_ele.attributes["src"])
```

oxinabox: Also when I look at that element in the webbrowser, I don't see a `src` attribute.
What is going on with that?
(It could be Github is actually showing us different sites, I guess)

oxinabox (changes requested):
Travis is being a bit weird right now.
It seems to be running the tests on the wrong commit.
We'll see it it fixes itself once you make the changes I suggest

src/GitHub.jl
SebastinSanty: GitHub has some sort of front-end rendering. If you go to `View Page Source (viewsource:)`, you'll able to see.

src/GitHub.jl
SebastinSanty: The regex shouldn't fail, but this is for the prevention in case of change in github's page structure like what happened previously with the current code. For me the master works, but I can revert it back if not desired.

oxinabox: fair enough, add a warning message here via `warn`,

oxinabox (approved):

SebastinSanty:
Any issues with merging this?

oxinabox:
Nope  no, issues, I was waiting for travis/appveyor to finish, when I approved it.
And then I forgot to come and check that it passed after.
Thanks for reminding me, sorry for losing track. (I wish github had a "Mark as unread")

PR #16, opened 2018-05-07
Adding build status badges

SebastinSanty:
DataDeps contains status badges so I thought it would be a good idea to include them for this project too.
* Need to update AppVeyor's badge's security token based badge image url


oxinabox:
Looks good.
I feel like `PkgDev` used to automatically include these in the template file

> Need to update AppVeyor's badge's security token based badge image url

What do you mean?
Seem to be working to me

SebastinSanty:
The AppVeyor badge which is there right now is taken from DataDeps.jl and
hence will reflect that package's status. I think only you will be able to
get the badge image for this package from your AppVeyor dashboard.

oxinabox:
Ah right, i get it.
Cool, I have grabbed that and fixed it.
Thanks.

PR #17, opened 2018-05-17
Add DataDeps Tests

SebastinSanty:
Not able to pass automatic "yes" to the tests


test/runtests.jl
```diff
@@ -4,7 +4,8 @@ using TestSetExtensions
 tests = [
     "UCI",
     "GitHub",
-    "DataDryad"    
+    "DataDryad",
+    "DataDepsTest"
```
oxinabox: Maybe "Integration Tests"  would be a better name?

SebastinSanty: Sure, I'll change it.

oxinabox (approved):
Looks good

oxinabox:
Looks like you need to regenerate one of the reference test files still

SebastinSanty:
The problem I am encountering is: I need silent "yes" to pass the `input_bool` question defined in DataDeps.jl. Silent yes as in the `-y` flag in `apt-get -y install something`. Do you know of how we can do that in Julia?

oxinabox:
Ah right,
DataDeps.jl has a config option (as an environment variable) to silently accept .

Add to the top of the integration tests file

```
ENV["DATADEPS_ALWAY_ACCEPT"]=true
```

Or set the environment variable in the appveryor and travis config

SebastinSanty:
Seems like DataDryad is down at the moment. @oxinabox Can you retrigger the Travis CI before you retire for the day?

oxinabox:
Ok, will do

oxinabox:
So this is only failing on AppVeyor, and it does so repeatably.
It works fine on my local machine (Windows 7)



The error code coming out is  (error: 2147500036)
This comes from windows urlmon (in julia 0.6.x)
In hex that is  0x80004004
Which is a  E_ABORT | Operation aborted
according to [windows docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa378137(v=vs.85).aspx)

According to the [docs](https://technet.microsoft.com/en-us/windows/ms775123(v=vs.60)) ULRMON shouldn't itself return that error code,
but I suspect there is any number of reasons why it might.

Now, the most reasonable reason for urlmon to give that back would be if 
it's callback had said to abort.
However, when julia uses urlmon it does not set that callback.
(see [here](https://github.com/JuliaLang/julia/blob/v0.6.2/base/interactiveutil.jl#L595))


Hmmmm...

##  First failure:
URL: https://datadryad.org/bitstream/handle/10255/dryad.173075/Fuzzy%20coded%20trait%20data.xlsx?sequence=1

Code: https://github.com/oxinabox/DataDepsGenerators.jl/pull/17/files#diff-594ffbb691d4b780125bfe4ea855f375R10

AppVeyor (Failing): https://ci.appveyor.com/project/oxinabox/datadepsgenerators-jl/build/1.0.48/job/e944rxte7d6u05y4#L308

TravisCI (working): https://travis-ci.org/oxinabox/DataDepsGenerators.jl/jobs/380857658#L869



## Second failure

URL: https://datadryad.org/bitstream/handle/10255/dryad.173251/raw_data.zip?sequence=1

Code:  https://github.com/oxinabox/DataDepsGenerators.jl/pull/17/files#diff-594ffbb691d4b780125bfe4ea855f375R11

AppVeyor (Failing):  https://ci.appveyor.com/project/oxinabox/datadepsgenerators-jl/build/1.0.47/job/73qwatndei0lja7v#L331

Travis CI (working): https://travis-ci.org/oxinabox/DataDepsGenerators.jl/jobs/380857658#L880

SebastinSanty:
@oxinabox Any replies from DataDryad or AppVeyor regarding this?

oxinabox:
Nothing since the tweak, I'll dm them again

oxinabox:
DataDryad says that are still looking into it

oxinabox:
DataDryad says:

> None of the IP addresses used by AppVeyor are blocked. Everything looks like it is responding fine on our end.

I'm going to see if it runs locally on my windows machine (I have to install julia).
And if so I'll then make a MWE and talk to AppVeyor support

oxinabox:
This is now passing, having restarted it again.
Must have been a transient fault.
Can you rebase it, and then we'll merge it

test/IntegrationTests.jl
```diff
@@ -0,0 +1,30 @@
+using DataDeps
+using Base.Test
+
+ENV["DATADEPS_ALWAY_ACCEPT"]=true
+
+@testset "Testing DataDyrad DataDeps" begin
+    include("references/DataDryad Drought.txt")
+    include("references/DataDryad Plasticity.txt")
```
oxinabox: Testing the copy that is in `"references" is not ideal.
Because if one makes a change, but forgets to update the reference copy,
then the change won't be reflected in the results.

I'm not sure the best way to avoid it.
One option is `eval(parse(generate(....)`.
(wrap that into a function)

test/REQUIRE
```diff
@@ -1,2 +1,3 @@
 ReferenceTests
 TestSetExtensions
+DataDeps
```
oxinabox, line 3: add MD5

test/IntegrationTests.jl
```diff
@@ -0,0 +1,30 @@
+using DataDeps
+using Base.Test
+
```
oxinabox: add `using MD5`

SebastinSanty:
Check #24

SebastinSanty:
Fixed by cad5daa

PR #18, opened 2018-05-18
Provision to add checksum in Register Blocks

SebastinSanty:


src/DataDepsGenerators.jl
```diff
@@ -71,6 +73,7 @@ function generate(repo::DataRepo,
         \"\"\"
     $(indent(message(meta)))\"\"\",
         $(meta.dataurls)
```
oxinabox: Comma seems to be missing here

src/DataDepsGenerators.jl
```diff
@@ -11,6 +11,7 @@ struct Metadata
     website::String
     description::String
     dataurls::Vector
+    datachecksums::Vector
```
oxinabox: make this unrestricted in type.
using `dataurls::Any`
because there are many options for a checksum.

 - `nothing`,   
 - `(md5, "234213542ababab23...")`,  
 - `"234213542ababab23..."`
 - Vectors of any combination of valid forms

src/GitHub.jl
```diff
@@ -70,6 +70,11 @@ function get_urls(repo::GitHub, page, cdn_url_converter=get_cdn_url_converter(pa
     urls
 end
 
+function get_checksums(repo::GitHub, page)
```
oxinabox: Rather than adding this code to every generator that doesn't have a way to get checksums,

add a fall back default method to the parent type of all Repos:
`DataRepo`

which would look like 

```
function get_checksums(repo::DataRepo, page)
     checksums = nothing #Nothing is the correct value to return, as it will trigger gr
     checksums
end
```

Or in short form
```
 get_checksums(repo::DataRepo, page) = nothing
```

That should go in the top level file

test/references/538 college.txt
```diff
@@ -453,5 +453,6 @@ RegisterDataDep(
...
+    Any["https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/all-ages.csv", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/college-majors-rscript.R", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/grad-students.csv", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/majors-list.csv", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/readme.md", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/recent-grads.csv", "https://cdn.rawgit.com/fivethirtyeight/data/ae8512edb504681c5e7251432507936d1064dfd7/college-majors/women-stem.csv"]
```
oxinabox: One day and that day is not today, we're going to remove the `Any`'s from this

oxinabox (commented):
Looks failing good.
A couple of points.

At some point I think some of the stuff for creating the actual text should move to its own file., rather than being in the top level file. Though that is not related to this particular PR

src/DataDryad.jl
```diff
@@ -49,6 +49,19 @@ function get_urls(repo::DataDryad, page)
     urls
 end
 
+function get_checksums(repo::DataDryad, page)
+    checksums = []
+    checksum_eles = matchall(sel"a", page.root)
+    for checksum_ele in checksum_eles
+        if ismatch(r"\bresource\/doi:[0-9]*.[0-9]*\/dryad.[a-z, 0-9]*\/[0-9]+\b", checksum_ele.attributes["href"])
+            checksum = match(r"\bresource\/doi:[0-9]*.[0-9]*\/dryad.[a-z, 0-9]*\/[0-9]+\b", checksum_ele.attributes["href"])
+            md5 = ("md5", text_only(getpage(replace(checksum.match, "resource/doi:", "https://datadryad.org/mn/checksum/doi:")).root))
```
oxinabox: Maybe this function can have emit a `info` that "The generated registration block uses `md5` hash, the MD5.jl package must be loaded to run the registration".

Since md5 is a function, but not one we've got inscope, (see discussion above about formatting it)
rather than making it a string, it would be nicer to use a Symbol,
written `:md5`.
Symbol is the type used in the AST to represent names of things.


Finally this line is overlong,
break it into 2 or 3

src/DataDepsGenerators.jl
```diff
@@ -70,11 +72,15 @@ function generate(repo::DataRepo,
         \"$shortname\",
         \"\"\"
     $(indent(message(meta)))\"\"\",
-        $(meta.dataurls)
+        $(meta.dataurls),
+        $(meta.datachecksums)
```
oxinabox: You can not simply directly interpolate in the datachecksums anymore.
Because 
The correct output is not:

```
Any[("md5", "0c484b24ed9b8ba6cfd85e6035b1a9e9"), ("md5", "8afefad6872124ceda8a2171465cf277"), ("md5", "91cff120293ae639d599b9057313f107")]
```

But is rather 

```
[(md5, "0c484b24ed9b8ba6cfd85e6035b1a9e9"), (md5, "8afefad6872124ceda8a2171465cf277"), (md5, "91cff120293ae639d599b9057313f107")]
```

So what I think you are going to need to do is do something like

`$(format_checksums(meta.datachecksums))`

and define something like:

```
format_checksums(csum) = cs # fallback

function format_checksums(csums::Vector)
	formattedlist = join(", ", format_checksums.(csums))
	"[$formattedlist]"
end

function format_checksums(csum::Tuple{T,<:AbstactString}) where T
	func, hashstring = cs
	"($func, \"$hashstring \")"
end
```
That is just an untested example
I will leave you to play with it and workout what is best

You could  make having `nothing` as the `meta.datachecksums` result in nothing at all being printed there

oxinabox (commented):
Sorry for my slowness in review, and that i didn't realize some of these things on my first pass.

Looking good though.
If you are not already, you're going to end up familiar with julia's (multiple) dispatch system which will be useful for the next stage

src/DataDepsGenerators.jl
```diff
@@ -70,11 +72,25 @@ function generate(repo::DataRepo,
...
 end
 
+function get_checksums(repo::DataRepo, page)
+    checksums = nothing
+    checksums
+end
+
+function format_checksums(csums::Vector)
+    formattedlist = csums[1]
+    for i in csums[2:end]
```
oxinabox: There is a Base Library function `join` for this

```
julia> join(["a", "b", "c"], ", ")
"a, b, c"
```

src/DataDryad.jl
```diff
@@ -49,6 +49,20 @@ function get_urls(repo::DataDryad, page)
...
+    for checksum_ele in checksum_eles
+        if ismatch(r"\bresource\/doi:[0-9]*.[0-9]*\/dryad.[a-z, 0-9]*\/[0-9]+\b", checksum_ele.attributes["href"])
+            checksum = match(r"\bresource\/doi:[0-9]*.[0-9]*\/dryad.[a-z, 0-9]*\/[0-9]+\b", checksum_ele.attributes["href"])
+            md5 = (:md5, text_only(getpage(replace(checksum.match, "resource/doi:", "https://datadryad.org/mn/checksum/doi:")).root))
+            info("The generated registration block uses md5 hash, the MD5.jl package must be loaded to run the registration")
+            push!(checksums, md5)
+        end
+    end
+    checksums
+end
```
oxinabox, line 67: I find this a bit hard to read.

Below is how I would have written this code:
Key points:

 1. Keep line length under 79 characters where possible
 2. Use more variables to express intermediary steps, (aids in 1, and in 3)
 3. Use meaningful variable names as it makes the code easier to understand, the things returned by ,`matchall(sel"a", page.root)`  is all links, not all checksums

For more specific points see comments


```
function get_checksums(repo::DataDryad, page)
    info("The generated registration block uses md5 hash, the MD5.jl package must be loaded to run the registration")
    # only print info once, it will apply to every checksum for DataDryad
    doi_regex = r"\bresource\/doi:[0-9]*.[0-9]*\/dryad.[a-z, 0-9]*\/[0-9]+\b"

    checksums = []
    link_eles = matchall(sel"a", page.root)
    for link_ele in link_eles
        doi_match = match(doi_regex,  link_eles.attributes["href"])
        if doi_match != nothing # match returns `nothing` on non-match
            doi = doi_match.match
            checksum_url = replace(doi, "resource/doi:", "https://datadryad.org/mn/checksum/doi:")
            md5_hash = text_only(getpage(checksum_url).root)
            md5 = (:md5, md5_hash)            
            push!(checksums, md5)
        end
    end
    @assert length(checksums)>0
    checksums
end
```

oxinabox (commented):
As well as the inline comments.

This is still not quiet right,
I am seeing `:md5` occur in the outputs,
but in the outputs it needs to just be `md5`


Below is the an testset for unit tests for `format_checksums`
Add it to the `/tests` dir and include it in `runtests`

```
import DataDepsGenerators:format_checksum
using Base.Test

@testset "Formatting Checksums" begin
    @test  format_checksum("898237b") ==
                    "\"898237b\""
                    
    @test  format_checksum(["898237b", "aba1"]) ==
                "[\"898237b\", \"aba1\"]"
                    
    @test  format_checksum((:md5, "898237b")) ==
                    "(md5,\"898237b\")"

    @test  format_checksum([(:md5, "898237b"),(:md6, "a1237b")]) ==
                "[(md5,\"898237b\"), (md6, \"a1237b\")]"
                
    @test  format_checksum([(:md5, "898237b"), "fafaf"]) ==
            "[(md5,\"898237b\"), \"fafaf\"]"
    

end
```

src/DataDepsGenerators.jl
SebastinSanty: I tried using `join()` but it was acting weird on joining list with tuples.

src/DataDryad.jl
SebastinSanty: Sorry, I forgot to heed to your review on reducing the line size. Right, I'll be more careful with variable names. Changing it.

SebastinSanty:
Thanks for the review! That testset gives me an idea of the checksums should be structured.

src/DataDepsGenerators.jl
oxinabox: What you want to be doing is converting all the tuples into their formatted strings before you join them.
So that you are only `join`ing strings

oxinabox:
The kind of solution I was hoping to have for `format_checksums`
was:

```
format_checksums(csum::AbstractString) = "\"$csum\""

function format_checksums(csum::Tuple{Symbol, <:AbstractString})
    func = string(csum[1])
    hashstring = format_checksums(csum[2])
    "($func, $hashstring)"
end

function format_checksums(csums::Vector)
    fcsums = join(format_checksums.(csums), ", ")
    "[$fcsums]"
end
```

Which is broadly similar to what you have.
The key idea that is missing is having `format_checksums` call itself, and letting it sort things via dispatch on type.
Rather than repeating every combination of options.

Also I'ld suggest being careful with how much you use `replace` to solve problems.
This package code I wrote is maybe is setting a bit of a bad example.
Direct string-manipulation is a bit of a hack.

src/DataDepsGenerators.jl
```diff
@@ -2,7 +2,7 @@ module DataDepsGenerators
 using Gumbo, Cascadia, AbstractTrees
 using Suppressor
 
-export generate, UCI, GitHub, DataDryad
+export generate, UCI, GitHub, DataDryad, format_checksums
```
oxinabox: No need to `export`, `format_checksums`

oxinabox (commented):
Other than the point I made about the nicer way to write `format_checksums` this is really close to done

SebastinSanty:
I am not sure how I can convert `:md5` to `md5` without using `replace()`. Am I missing something?

oxinabox:
yep,
the `string` function. (Not to be confused with the `String` constructor)

```
julia> x = :md5
:md5

julia> typeof(x)                                                                                                                        Symbol

julia> string(x)
"md5"

julia> string(:anything)
"anything"
```

oxinabox:
Broadly speaking looking good now: lets see what CI says 

For your education  I suggest reading and understanding 
https://docs.julialang.org/en/stable/manual/functions/#man-vectorized-1
I where possible, I prefer to write `foo.(xs)` over `map(foo, xs)` or `[foo(x) for x in xs]`
though it is mostly a stylistic choice.

Also can you explain to me (and perhaps yourself) why
you wrote:
`format_checksums(csum::Tuple{T,<:AbstractString}) where T<:Symbol`

and not
`format_checksums(csum::Tuple{Symbol,<:AbstractString})`

It isn't wrong as such.
But it is worth understanding the difference between `Foo(x::Vector{Bar})`  vs `foo(x::Vector{T}) where T<: Bar`  (also written `foo(x::Vector{<:Bar})`
When they are the same and when they are different

SebastinSanty:
From what I understand, `Foo(x::Vector{Bar})` will strictly work for `Bar` type arguments (in the Vector) whereas `Foo(x::Vector{<:Bar})` will work for arguments of any type (in the Vector) which comes under the Bar hierarchy.

oxinabox:
Cool, you've got that basically right.
So the next trick is that we have abstract types (like `AbstractString`, `Real`, `AbstractArray` etc),
and concrete types (like `String`, `Float32`, `Array`).
Only abstract types can have subtypes.

`Symbol` is a concrete type, so no subtypes.
Thus the only thing matched by `T<:Symbol` is `Symbol`.
So  rather than ever saying `Foo(::Vector{<:Symbol})` one can just say `Foo(::Vector{Symbol})`.

SebastinSanty:
Should I be changing it?

oxinabox:
It doesn't really matter to me if you change it in this PR or just catch it later, or leave it

oxinabox:
Merged :tada:

PR #19, opened 2018-05-24
Implement DataDryadAPI

SebastinSanty:
I have a few points to make:

* While I was coding up, it was becoming increasing difficult to know what is to be kept in the purview of DataOne and what is to be kept in DataDryad. My thoughts on this were to go ahead with `DataDryadAPI <: DataRepo`, and the when adding support for the Arctic Data Center repos (in a new PR), implement the common DataOne base, such that it will become:
```
DataDryad <: DataOne <: DataRepo
ArcticDataCenter<: DataOne <: DataRepo
```
* Regarding citation format, I was not able to find one on the API. So I manually constructed one (which is obviously not the right way to do as stuff like journal name is missing)
* License is the same for all DataDryad repos, so I kept it as a constant. Anyways, I was not able to retrieve that from the API.
* Bitstream urls on DataDryad which is used for downloading datasets and which we had an issue with earlier, is still not working. Hence, I am not putting the bitstream urls but instead parsing the same old way.
* Should I be providing checksums for each download? I didn’t implement it in the DataDryadWeb, but it is now there in this PR.
* I have a strong urge to start segregating the source/tests/references in folders for better understandability.


src/DataDryadAPI.jl
```diff
@@ -0,0 +1,67 @@
...
+function description(repo::DataDryadAPI, mainpage)
+    desc = text_only(first(matchall(sel"dcterms\:description", mainpage.root)))
+    authors = matchall(sel"dcterms\:creator", mainpage.root)
+    author = text_only(first(authors))
+    license = "http://creativecommons.org/publicdomain/zero/1.0/"
+    rawdate = chop(text_only(first(matchall(sel"dcterms\:dateSubmitted", mainpage.root))))
+    year = Dates.year(Dates.DateTime(rawdate))
+    month = Dates.monthname(Dates.DateTime(rawdate))
+    day = Dates.day(Dates.DateTime(rawdate))
+    date = "$(month) $(day), $(year)"
```
oxinabox: You can format this a bit more cleanly using `date.format`
https://docs.julialang.org/en/stable/manual/dates/

Something like
```
date = Dates.DateTime(rawdate)
Dates.format(dt, "mm dd, yyyy")
year = Dates.year(date)
```

I think

src/DataDryadAPI.jl
```diff
@@ -0,0 +1,67 @@
+immutable DataDryadAPI <: DataRepo
+end
+
+base_url(::DataDryadAPI) = "https://datadryad.org/mn/object/http://dx.doi.org/"
+
+function description(repo::DataDryadAPI, mainpage)
+    desc = text_only(first(matchall(sel"dcterms\:description", mainpage.root)))
+    authors = matchall(sel"dcterms\:creator", mainpage.root)
+    author = text_only(first(authors))
```
oxinabox: rather than doing this directly, each time 
I suggest you create a general function, call it `short_author(authors)`  which takes in the full author list,
and if only 1 returns just that one.
If 2 returns the first and the second
and if 3 + returns first et al.

Maybe using dispatch and splatting is a good way: 
That would look like

```
short_author(author1) = author1
short_author(author1, author2) = author1 * " and " * author2
short_author(author1, authors...) = author1 * " et al."
```

Invoking it as `short_author(authors...)` to splat the argument


Also note it is "et al." , with out the period after the "et"

test/references/DataDryadAPI Drought.txt
```diff
@@ -0,0 +1,21 @@
+RegisterDataDep(
+    "Data from Drought intensification drives turnover of structure and function in stream invertebrate communities",
+    """
+	Dataset: Data from: Drought intensification drives turnover of structure and function in stream invertebrate communities
+	Website: https://datadryad.org/mn/object/http://dx.doi.org/10.5061/dryad.cc8834s
```
oxinabox: This URL is wrong
The URL you should be linking to is
http://dx.doi.org/10.5061/dryad.cc8834s

oxinabox (commented):
Looking generally pretty good

oxinabox:
>  While I was coding up, it was becoming increasing difficult to know what is to be kept in the purview of DataOne and what is to be kept in DataDryad. My thoughts on this were to go ahead with DataDryadAPI <: DataRepo, and the when adding support for the Arctic Data Center repos (in a new PR), implement the common DataOne base, such that it will become: `DataDryad <: DataOne <: DataRepo`, `ArcticDataCenter<: DataOne <: DataRepo`

I agree this is an excellent strategy, avoiding premature (and potentially incorrect) abstraction.

>    Regarding citation format, I was not able to find one on the API. So I manually constructed one (which is obviously not the right way to do as stuff like journal name is missing)

It is good enough for now.
Once we move on doing more stuff with DOIs directly, e.g. relating to DataCite,
we can look at some of the solutions in that space that automatically generate formatted citations.

> License is the same for all DataDryad repos, so I kept it as a constant. Anyways, I was not able to retrieve that from the API.

Reasonable

>   Bitstream urls on DataDryad which is used for downloading datasets and which we had an issue with earlier, is still not working. Hence, I am not putting the bitstream urls but instead parsing the same old way.

I will chase this up with their support

>   Should I be providing checksums for each download? I didn’t implement it in the DataDryadWeb, but it is now there in this PR.

I'm not sure I understand the question, AFAICT you did provide checksums for each download in 
DataDryadWeb
https://github.com/oxinabox/DataDepsGenerators.jl/blob/883457f7944f4aa4febb8361561332c0da211a4b/test/references/DataDryad%20Drought.txt#L19-L20
In general there isn't a consistently defined notion for a checksum for multiple files.
(DataDeps.jl implements one, but it is only valid for checksums it has generated in the first place)

>    I have a strong urge to start segregating the source/tests/references in folders for better understandability.

Sounds reasonable. Don't go too overboard

oxinabox:
That is looking pretty good to me.
If your done, I'll merge after appveyor completes

PR #21, opened 2018-05-25
Implement DataOneV2 abstraction

SebastinSanty:
Few points to make:

* It happens that DataDryad is on a older version of DataOne API, hence what was described yesterday didn’t quite turn out as expected. Hence, I made a common base for DataOnev2 which includes KNB which in turn supports ArcticDataCenter. I was not able to find any other API (I was trying to get TERN AU, but didn’t work), hence it would be very helpful if you can point me to a few. Also, Merritt by UC3 which is supposed to be on DataOne API supports a completely different API.

* Though ArcticDataCenter(ADC) mentions that it runs its API through KNB (check [here](https://arcticdata.io/catalog/#api)), I was not able to get ADC examples working with its API other than the one example mentioned in that doc.


src/DataDepsGenerators.jl
```diff
@@ -101,8 +102,10 @@ function format_authors(authors::Vector)
         authors[1]
     elseif length(authors) == 2
         authors[1] * ", " * authors[2]
-    else
+    elseif length(authors) >2
         authors[1] * " et al."
+    else
```
oxinabox: Maybe make this spit out a warning to the console too

src/DataDepsGenerators.jl
```diff
@@ -101,8 +102,10 @@ function format_authors(authors::Vector)
...
+        "Unknown Author"
```
oxinabox: Maybe make this spit out a warning to the console too

src/DataOnev2/DataOnev2.jl
```diff
@@ -0,0 +1,34 @@
...
+
+function description(repo::DataOnev2, mainpage)
+    desc = try text_only(matchall(sel"description para", mainpage.root)) catch "" end
+    authors = try [text_only(i) for i in matchall(sel"individualName", mainpage.root) if contains(string(i.parent), "creator")] catch [] end
+    author = format_authors(authors)
+    license = try text_only(first(matchall(sel"intellectualRights", mainpage.root))) catch "" end
+    rawdate = try Dates.Date((text_only(first(matchall(sel"pubDate", mainpage.root)))), "yyyy-mm-dd") catch "" end
+    year = try Dates.year(rawdate) catch "" end
+    date = try Dates.format(rawdate, "U d, yyyy") catch "" end
+    references = ""
```
oxinabox: Please no single line `try`, `catch` blocks.
And especially no catch all conditions.
So called Pokemon-Exception handleing
http://wiki.c2.com/?PokemonExceptionHandling


try block should should only really be used with the following pattern:



```
try
    # code
catch err
   if err isa SpecificExceptionType # specify exception type here
       # code to handle exception
   else
         rethrow()
   end
end
```

It is much better to know what if failing and why,
and then handle it correctly.
than it is to catch everything and produce a result that is silently wrong without explanations.

In general one also would want to spit out warning is a lot of these failures even if they can be handled.

src/DataOnev2/DataOnev2.jl
```diff
@@ -0,0 +1,34 @@
...
+    paper = join(authors, ", ") * " ($(year)) " * data_fullname(repo, mainpage) * " " * references
```
oxinabox: This does not give you the name of the paper.
If one can't get the name of the paper, it is fine to exclude it from the generation.
If the DataOne API doesn't keep that kinda metadata then it is not a worry.

src/DataOnev2/KNB.jl
```diff
@@ -0,0 +1,19 @@
+abstract type KNB <: DataOnev2 end
+
+base_url(::KNB) = "https://knb.ecoinformatics.org/knb/d1/mn/v2/"
```
oxinabox: You want to define:
`object_url(repo) = base_url(repo) * "object/"`
or similar 

Then remove a bunch of string literals from your code:
`"https://knb.ecoinformatics.org/knb/d1/mn/v2/object/"`

This follows the DRY principle https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

oxinabox (commented):
having files just to contain a single type with no methods is not useful.
Just put them in the file with the parent type that actually has methods

src/DataOnev2/DataOnev2.jl
SebastinSanty: But this is for the citation. I was already creating this in the previous PR, and as you had termed it is okay, I considered putting it here too.

src/DataOnev2/DataOnev2.jl
SebastinSanty: I knew this was coming for writing bad code :/. The only issue is that, this being at an abstraction layer, some of the repo APIs don't have the tags required. I felt `try` was inevitable, but sure I'll catch the exceptions in a better way. Keeping into account your previous comment about files with single type, I think to avoid `try`, I need to write some logic (which removes the tags not present) in those files.

oxinabox: maybe something like `author(repo, page)`, and `data(repo, page)`

src/DataOnev2/DataOnev2.jl
oxinabox: Ah, sorry I messed up in my review there.
I didn't understand.

I think best course is to just skip producing anything about what to cite,
(or maybe just stick the DOI numbers there)

Until we can do something like
I suggest in #22

oxinabox:
I suggest the capitalization be `DataOneV2` not `DataOnev2`

src/DataOnev2/DataOnev2.jl
```diff
@@ -21,14 +19,49 @@ function description(repo::DataOnev2, mainpage)
...
 
-get_urls(repo::DataOnev2, page) = []
+get_urls(repo::DataOneV2, page) = []
 
-function data_fullname(::DataOnev2, mainpage)
+function data_fullname(::DataOneV2, mainpage)
     text_only(first(matchall(sel"description title", mainpage.root)))
+end
+
+function pubDate(repo::DataOneV2, mainpage)
```
oxinabox: Normal convention is `snake_case`, not `camelCase` for function names.

src/DataOnev2/DataOnev2.jl
```diff
@@ -0,0 +1,67 @@
...
+    """, "\$")
+end
+
+get_urls(repo::DataOneV2, page) = []
+
+function data_fullname(::DataOneV2, mainpage)
+    text_only(first(matchall(sel"description title", mainpage.root)))
+end
+
+function pubDate(repo::DataOneV2, mainpage)
```
oxinabox: convention is `snake_case` not `camelCase` for function names

src/DataOneV2/TERN.jl
```diff
@@ -0,0 +1,30 @@
+immutable TERN <: DataOneV2 end
```
oxinabox: use `struct` rather than `immutable`.

src/DataOneV2/TERN.jl
```diff
@@ -0,0 +1,30 @@
...
+
+base_url(::TERN) = "https://dataone.tern.org.au/mn/v2/object/aekos.org.au/collection/nsw.gov.au/nsw_atlas/"
+
+function get_urls(repo::TERN, page)
+    urls = []
+    links = matchall(sel"distribution online url", page.root)
+    push!(urls, text_only(links))
+    urls
```
oxinabox: Why not write:
```
function get_urls(repo::TERN, page)
     links = matchall(sel"distribution online url", page.root)
     text_only(links))
end
```
?

src/DataOneV2/TERN.jl
```diff
@@ -0,0 +1,30 @@
...
+    [text_only(i) for i in author_ele]
+end
+
+function pub_date(repo::TERN, mainpage)
+    date_ele = matchall(sel"temporalCoverage rangeOfDates endDate", mainpage.root)
+    rawdate = Dates.Date(text_only(first(date_ele)), "yyyy-mm-dd")
+    Dates.year(rawdate), Dates.format(rawdate, "U d, yyyy")
+end
+
+function data_fullname(::DataOneV2, mainpage)
```
oxinabox: `data_fullname(::TERN, mainpage)`
I suspect you mean

oxinabox (approved):
Looks good.
I am not too worried about the app-veyor failure when travis is working.
Probably just flacky network stuff

PR #24, opened 2018-06-02
 Add Integration Tests

SebastinSanty:


oxinabox:
What do you think about the idea putting the integration tests in with the reference tests?

E.g.

```
@testset "BuzzFeedNews (whole repos)" begin
	
	@testset "Pres Camp" begin
		registration_code = generate(GitHub(), "https://github.com/BuzzFeedNews/presidential-campaign-contributions")
			
		@testset "Integration Test" begin 
			eval(parse(registration_code)) # evaluate the new code
			@test length(collect(readdir(datadep"Presidential Campaign Contributions"))) > 0
		end
		
		@test_reference "references/buzzfeed pres-camp.txt" registration_code # See if new code is same as old
	end	
	
    	
    @test_reference "references/buzzfeed primates.txt" generate(GitHub(), "BuzzFeedNews/2015-07-primates")
end
```


Advantages I see is: 

 - it keeps it localized to anything that changed.
 - It means it is easy to test the newly generated code against running correctly (integration tests) &  matching old code (reference tests)
 - Right now only the reference results have integration tests. Which means if the reference tests fail the integration test results are meaningly (maybe even misleading) since they are running on the old code.

SebastinSanty:
Yes, right. That'll be great. I'll make the modifications.

SebastinSanty:
@oxinabox Any comment on this? I have changed it according to your suggestion. Also I am not very sure about how to set environment variables in AppVeyor, as it is not covered much in their documentation. Should I rather define the env var in all the files than in the CI yml file?

oxinabox:
What you've done looks right
https://www.appveyor.com/docs/build-configuration/#custom-environment-variables

test/DataDryadAPI.jl
```diff
@@ -4,6 +4,16 @@ using Base.Test
 using ReferenceTests
 
 @testset "DataDryadAPI test" begin
+    # @testset "DataDryadWeb Ecology" begin
```
oxinabox, line 7: Why is this one commented out?

oxinabox (dismissed):
Other than the fact that one is commented out, without anything explaing why.
This looked good

oxinabox:
I have just pushed a commit to your branch which I think fixes A.V config

test/DataDryadAPI.jl
SebastinSanty: For some weird reason, I am able to manually download the files specified in the urls, but the test are not being able to do so. I'll have to try again.

oxinabox: Is that everywhere not able to or just A.V being weird.

SebastinSanty: Locally too. I don't remember the exact error, I'll try and revert back to you.

oxinabox:
The A.V failure is the usual one I think

oxinabox (approved):

test/DataDryadAPI.jl
SebastinSanty:
```
Integration Test: Test Failed
  Expression: length(collect(readdir(@datadep_str("Data from Ecology and genomics of an important crop wild relative as a prelude to agricultural innovation")))) > 0
   Evaluated: 0 > 0
```

PR #26, opened 2018-06-06
Add CKAN API

SebastinSanty:


src/CKAN.jl
```diff
@@ -0,0 +1,56 @@
+using JSON
```
oxinabox: `using` statements are best put only in the top-level file. (`src/DataDepsGenerators.jl`).
I speak from experience when I say putting them everywhere leads to a mess

src/CKAN.jl
```diff
@@ -0,0 +1,56 @@
...
+
+immutable CKAN <: DataRepo
```
oxinabox: use `struct` rather than `immutable`.

src/CKAN.jl
```diff
@@ -0,0 +1,56 @@
...
+end
+
+base_url(::CKAN) = "https://datadryad.org/mn/object/http://dx.doi.org/"
```
oxinabox: This can't be right, can it? I don' tthink that is the base URL for every (or even any) CKAN page.
If it isn't being hit, delete the method

src/CKAN.jl
```diff
@@ -0,0 +1,56 @@
...
+
+function description(repo::CKAN, mainpage)
+    j = JSON.parse(text_only(mainpage.root))
```
oxinabox: No single letter variables.
They are impossible to ctrl+f for.
I'd call this `data` or something

oxinabox: Probably even

```
data = JSON.parse(text_only(mainpage.root))["result"]
```

So you can avoind indexing into `result` each time it is used below

src/CKAN.jl
```diff
@@ -0,0 +1,56 @@
...
+    desc = j["result"]["notes"]
+    authors = [j["result"]["notes"]]
```
oxinabox: Should be:

```
 authors = [j["result"]["author"]]
```

Also CKAN has a maintainer field,
that should probably be included:
`j[`results`][`maintainer`]

It also has fields for the email address for both of them, which might also be good to have.

test/CKAN.jl
```diff
@@ -0,0 +1,11 @@
+using DataDepsGenerators
+using Base.Test
+
+using ReferenceTests
+
+@testset "CKAN test" begin
+    @test_reference "references/CKAN Gold Prices.txt" generate(CKAN(), "https://demo.ckan.org/api/3/action/package_show?id=gold-prices")
```
oxinabox: Ideally,
I think the better interface would be to be calling:

```
@test_reference "references/CKAN Gold Prices.txt" generate(CKAN(), "https://demo.ckan.org/dataset/gold-prices")
````

and then transforming that URL into "https://demo.ckan.org/api/3/action/package_show?id=gold-prices" as part of the package.

I have a feeling the `find_metadata`  function needs reengineering.
Nowish is probably a good time to do that kind of refactor.

**Review by oxinabox — COMMENTED** [review]
It is unfortunate that this requires reparsing the main page several times.

How should we refactor the package to avoid this?

**Comment by oxinabox** [comment]
Do you know the type of the hash?
I can't work it out

https://stackoverflow.com/q/50712671/179081

**Comment by SebastinSanty** [comment]
No, I am not able to figure out which type of hash it is :/

**Comment by oxinabox** [comment]
The big question to me is 
how do we go from a landing-page on a site that supports CKAN,  to a CKAN, API page.

<s>Right now this only works for the CKAN demo site right? </s>


Examples:

 - Data.gov.au
    - Landing page: https://data.gov.au/dataset/2016-soe-atmosphere-hourly-co-and-24h-pm2-5-concentrations-measured-during-the-hazelwood-mine-fire
    - API page: https://data.gov.au/api/3/action/package_show?id=2016-soe-atmosphere-hourly-co-and-24h-pm2-5-concentrations-measured-during-the-hazelwood-mine-fire
 
 - Data.gov
    - Landing Page: https://catalog.data.gov/dataset/consumer-complaint-database
    - API page: https://catalog.data.gov/api/3/action/package_show?id=consumer-complaint-database

Maybe the solution is to just replace the first use if the word `dataset` in the URL with the string `api/3/action/package_show?id=`

Which is roughly what you are doing.

It seems simple and fragile, but if the whole site is actually built onto of CKAN maybe it will always have that structure.

File: src/CKAN.jl [hunk 7]
```diff
@@ -0,0 +1,60 @@
+struct CKAN <: DataRepo
+end
+
+base_url(::CKAN) = "https://demo.ckan.org/api/3/action/package_show?id="
```
oxinabox: I don't think this is useful, given there are many sites that use this,
There is no single canonical `base_url`.

src/CKAN.jl
```diff
@@ -0,0 +1,60 @@
...
+
+function website(::CKAN, mainpage_url)
+    replace(mainpage_url, "/api/3/action/package_show?id=", "/dataset/")
+end
+
+function mainpage_url(repo::CKAN, dataname)
+    if startswith(dataname, "http")
+        url = replace(dataname, "/dataset/", "/api/3/action/package_show?id=")
+    else # not a URL
+        url = base_url(repo) * dataname
```
oxinabox: I think better to just throw an error here.
We don't know what site is intended with only an ID.
and hopefully for CKAN we can avoid needed to implement per-site solutions

test/references/CKAN District Spending.txt
```diff
@@ -0,0 +1,21 @@
+register(DataDep(
+    "UK Adur District Council Spending Data",
+    """
+	Dataset: UK: Adur District Council Spending Data
+	Website: https://demo.ckan.org/dataset/adur_district_spending
+	Author: Lucy Chambers,  (Maintainer)
+	License: Creative Commons Attribution Share-Alike
+	Date: July 18, 2017
+	
+	From Spikes Cavell, Spotlight on Spend.\\r
```
oxinabox: Hmmm, it is not great that `\\r` is showing up in the output.
That is likely the result of the call to `escape_multiline_string`,
as that is a doubly escaped carriage return.
It is odd that it is showing up, I'ld never noticed it before.

I don;t think needs to be fixed in this PR.
Might be good to open an issue to track it through

src/CKAN.jl
```diff
@@ -0,0 +1,60 @@
+struct CKAN <: DataRepo
+end
+
+base_url(::CKAN) = "https://demo.ckan.org/api/3/action/package_show?id="
+
+function description(repo::CKAN, mainpage)
+    desc = mainpage["notes"]
+    authors = [mainpage["author"], mainpage["maintainer"] * " (Maintainer)"]
+    author = format_authors(authors)
```
oxinabox: Author and maintainer should be different fields.
The maintainer potentially was not involved with the original data creation. 
Best to just list both seperately

src/CKAN.jl
```diff
@@ -0,0 +1,60 @@
...
+    dataset = mainpage["name"]
+    
+    final = escape_multiline_string("""
+    Author: $(author)
+    License: $(license)
+    Date: $(date)
+
+    $(desc)
+
+    Please cite this dataset:
```
oxinabox: i would remove the `Please cite` section from CKAN, it really belong to data that is linked to accademic research.
CKAN is for goverment publications, and the conventions are different.
There is no paper to cite and the datasets don't e.g. have a DOI number.
The grey literature people probably have something to say about it, but for now it is beyond the scope of the project I think.
They top matter with the info is enough

test/CKAN.jl
```diff
@@ -0,0 +1,11 @@
+using DataDepsGenerators
+using Base.Test
+
+using ReferenceTests
+
+@testset "CKAN test" begin
+    @test_reference "references/CKAN Gold Prices.txt" generate(CKAN(), "https://demo.ckan.org/dataset/gold-prices")
+    
+    #Checking files with multiple files available for download
+    @test_reference "references/CKAN District Spending.txt" generate(CKAN(), "adur_district_spending")
```
oxinabox: This line will need to be updated if BaseURL is removed

src/CKAN.jl
```diff
@@ -0,0 +1,60 @@
...
+    for i = 1:page["num_resources"]
+        push!(urls, page["resources"][i]["url"])
+    end
+    urls
+end
+
+function get_checksums(repo::CKAN, page)
+    checksums = []
+    for i = 1:page["num_resources"]
+        if page["resources"][i]["hash"] != "" push!(checksums, page["resources"][i]["hash"]) end
```
oxinabox: Please no one-line if statements.
Too hard to read.

It might be with this outputting an `info("Checksum was found, however CKAN checksums are of unknown type, we are assuming SHA256")`

oxinabox (commented):
Looking pretty good.
Can you add tests based on the two links I gave a abot (or other ones from Data.gov and Data,gov.au)

And also add integration tests.

BTW
Do you know what the difference between the `title` and the `name` fields are?

Some I was looking at had both

oxinabox:
Got an answer on stack overflow:
https://stackoverflow.com/a/50758819/179081

So looks like the hash isn't consistent between upload tools.
And I'm guessing we can't tell what the upload tool was.

So I think best would be to just not emit the hash,
and fall back to DataDeps.jl's default behavor where it tells  the user the hash of the file on the first download so they can edit it into their source. 

We could think about making PR's to all the clients (Line numbers are linked in the S.O post)
but that would be a long term solution.

oxinabox:
Other thing is we need integration tests.
After that, I think can merge.

It might be worth defining a macro or function that triggers the integration and reference tests and just using that everywhere.

Does generate take a datadep name as an argument? It should.

Then can do something like

```
function test_generation(repo, url, name)
	@testset "name" begin
		registration = generate(repo, url; name=name)
		@reference_test "references/$name" registration
		
		eval(parse(registration))
		@test length(readdir(@datadep_str(name))) > 0 
	end
end
```

Might be better to do that as another PR, another time.

src/CKAN.jl
```diff
@@ -30,14 +30,7 @@ function get_urls(repo::CKAN, page)
...
 function get_checksums(repo::CKAN, page)
-    checksums = []
-    for i = 1:page["num_resources"]
-        if page["resources"][i]["hash"] != "" 
-            push!(checksums, page["resources"][i]["hash"])
-        end
-    end
-    info("Checksum was found, however CKAN checksums are of unknown type, we are assuming SHA256")
-    checksums
+    []
```
oxinabox: `nothing` is the return you want.
Because you don't want to generate `[]` in the output.

oxinabox (approved):

PR #31, opened 2018-06-18
Fix GitHub broken tests

SebastinSanty:


SebastinSanty:
Need to merge #27 before this.

SebastinSanty:
@oxinabox This can be merged, I guess.

oxinabox:
Cool thanks. Nice work.
At some point we'll maybe have to pull out all the github webscraping stuff and use the Github API directly.
I can't remember why I didn't do that in the first place.
Though I feel like I had a reason.
(Maybe I was wanting to download full zip/tarballs of Repos, but never implemented it.)

Issue #30, opened 2018-06-15
Use  linked data to handle (nearly) arbitrary websites?


oxinabox:
I wonder if the deployment of linked data solutions is widespreed enough that we can use that
for arbitrary URLs.
Linked data formats like [JSON-LD](https://en.wikipedia.org/wiki/JSON-LD) are widely deployed because search engines etc like them.

Not evertyhing has it, for example Github for example doesn't have a JSON-LD.

## Figshare seems to include JSON-LD on every page:

This has everything we want including a download URL  (except hashes), and 

## eg1
```javascript
lyndon@agent:~$ curl -LH "Accept: text/html" https://doi.org/10.6084/m9.figshare.5350216.v1   2> /dev/null | xmllint -nowarning --html --xpath '//script[@type="application/ld+json"]' - 2> /dev/null
<script type="application/ld+json">{
    "@context": "http://schema.org",
    "@type": "Dataset",
    "name": "figshare for Institutions - Information booklet",
    "description": "figshare for Institutions information booklet about managing and disseminating research data to make it more citable, shareable and discoverable.",
    "url": "https://figshare.com/articles/figshare_for_Institutions_-_Information_booklet/5350216",
    "sameAs": "https://figshare.com/articles/figshare_for_Institutions_-_Information_booklet/5350216",
    "version": "1",
    "keywords": "figshare, figshare for Institutions, data management, research management, research data, discoverability, FFI, Japan, September 2017, openscijapan2017",
    "variablesMeasured": "none",
    "creator": [

      {
        "@type": "Person",
        "name": "figshare figshare"
      }
    ],
    "distribution": [
      {
        "@type": "DataDownload",
        "contentUrl": "https://ndownloader.figshare.com/files/9194386",
        "license": "https://creativecommons.org/licenses/by/4.0/"
      }
    ]
  }
```
## eg 2

```javascript
lyndon@agent:~$ curl -LH "Accept: text/html" https://figshare.com/articles/_shows_examples_of_coordinated_and_uncoordinated_motion_for_dangerous_and_non_dangerous_crowd_behavior_/186003  2> /dev/null | xmllint -nowarning --html --xpath '//script[@type="application/ld+json"]' - 2> /dev/null

<script type="application/ld+json">{
    "@context": "http://schema.org",
    "@type": "Dataset",
    "name": "shows examples of coordinated and uncoordinated motion for dangerous and non-dangerous crowd behavior.",
[...]

PR #34, opened 2018-06-22
Add Figshare API

SebastinSanty:


src/DataDepsGenerators.jl
```diff
@@ -115,11 +116,15 @@ function format_papers(authors::Vector, year::String, name::String, link::String
...
 
 function match_doi(uri::String)
-    identifier = match(r"\b(10[.][0-9]{4,}(?:[.][0-9]+)*\/(?:(?![\"&\'<>])\S)+)\b", uri).match
-    return identifier
+    try
+        identifier = match(r"\b(10[.][0-9]{4,}(?:[.][0-9]+)*\/(?:(?![\"&\'<>])\S)+)\b", uri).match
+        return true, identifier
+    catch ErrorException
+        return false
+    end
```
oxinabox: This is not a nice return value.
`Union{Void, T}` is a standard pattern for return values in julia. (and is becoming more so in 0.7).

`Union{Bool, Tuple{Bool, T}}` is not.

What is throwing an `ErrorException` here? Most things in `Base` should not throw that kind of exception, so it might be worth raising up stream

src/Figshare.jl
```diff
@@ -0,0 +1,74 @@
...
+    end
+    urls
+end
+
+function get_checksums(repo::Figshare, page)
+    checksums = []
+    for i in page["files"]
+        push!(checksums, i["computed_md5"])
+    end
+    checksums
```
oxinabox: Need to do the same thing as for the other that is MD5,
give a message telling user it is MD5 and setting it so that `(md5, "1234A12")` is produced in the output.
Could abstract out a common unction for that.

src/Figshare.jl
```diff
@@ -0,0 +1,74 @@
...
+    mainpage["url_public_html"]
+end
+
+function match_figshare(uri::String)
+    try
+        identifier = match(r"\d{7}", uri).match
+        return true, identifier
+    catch ErrorException
+        return false
+    end
```
oxinabox: See above. This is not a nice return type.

src/Figshare.jl
```diff
@@ -0,0 +1,74 @@
...
+    if match_doi(dataname)[1]
+        identifier = match_doi(dataname)[2]
+        down_url = base_url(repo) * "?doi=" * identifier
+        doi_page = JSON.parse(text_only(getpage(down_url).root))
+        url = doi_page[1]["url_public_api"]
+    elseif match_figshare(dataname)[1]
+        identifier = match_figshare(dataname)[2]
+        url = base_url(repo) * "/" *identifier   
+    else
+        error("Please use a valid url")
```
oxinabox: You mean `error("Please use a valid url, DOI, or Figshare ID")`

oxinabox (commented):
This is mostly right.
A few things still need to be fixed.

I take it this being done via Figshare API and not OAI-PMH means that OAI-PMH is not suitable for our purposes?

With that said: while this works
Does it give anything over 
doing 
#30.
Which would cover many sites including: 
Figshare, Kaggle, and Zenodo 
and partially cover DataVerse sites
?

I'm not sure if it is better to finish this one off,
Then look at #30 later to get the others,
or switch to doing something based on #30 now.

Broader picture do you have  any thoughts on restructuring the project more generally?
It might be good to spend time paying down some tech debt.

src/Figshare.jl
oxinabox: Wait how did it pass tests locally if it was not using MD5 to check MD5s?

src/DataDepsGenerators.jl
SebastinSanty: The `ErrorException` is to handle `Void` (in case of no regex match) which doesn't have a match attribute. What would you suggest to do otherwise, in case of returning values. I can do `try catch` instead of `if else` but I don't think its preferred in any way.

src/Figshare.jl
SebastinSanty: Yes, I'll update it.

src/Figshare.jl
SebastinSanty: Because we don't have Integration Tests for others using MD5.

SebastinSanty:
Yes, this is being done using Figshare API. I thought of doing OAI-PMH afterwards as you were more willing to get Figshare and Dataverse integrated.

My thoughts are: #30 in general gives amazing results but, have various attributes missing. However all of them mention the source of the data explicitly. What if we have individual types like `Figshare()` and `Dataverse()` and a general `JSONLD()` which eventually gets the JSON-LD, parses the source, and passes it into one of the types which we have already integrated.

src/Figshare.jl
```diff
@@ -0,0 +1,74 @@
...
+        identifier = match_doi(dataname)[2]
+        down_url = base_url(repo) * "?doi=" * identifier
+        doi_page = JSON.parse(text_only(getpage(down_url).root))
+        url = doi_page[1]["url_public_api"]
+    elseif match_figshare(dataname)[1]
+        identifier = match_figshare(dataname)[2]
+        url = base_url(repo) * "/" *identifier   
+    else
+        error("Please use a valid url, DOI, or Figshare ID")
+    end
```
oxinabox: How about leave `match_doi` returning just `nothing` or the DOI.
And handle it with dispatch (the julian way :-D)

Not tested but the following I think should work.
It define 2 local functions, each with 2 methods.
One for `::Void` and one for `::RegexMatch`

```
function mainpage_url(repo::Figshare, dataname)
	get_url_doi(::Void) = get_url_figshare(match_figshare(dataname))
	function get_url_doi(identifier_match::RegexMatch)
		identifier = identifier_match.match
        down_url = base_url(repo) * "?doi=" * identifier
        doi_page = JSON.parse(text_only(getpage(down_url).root))
		url = doi_page[1]["url_public_api"]
		return url
	end
	
	get_url_figshare(::Void) = error("Please use a valid url, DOI, or Figshare ID")
	function get_url_figshare(identifier_match::RegexMatch)
		identifier = identifier_match.match
		url = base_url(repo) * "/" *identifier
		return url
	end
	
        get_doi_url(match_doi(dataname))
end
```

You could also write that without using dispatch to local functions:

```
function mainpage_url(repo::Figshare, dataname)
	identifier_match = match_doi(dataname)
	if identifier_match isa RegexMatch
	    identifier = identifier_match.match
        down_url = base_url(repo) * "?doi=" * identifier
        doi_page = JSON.parse(text_only(getpage(down_url).root))
		url = doi_page[1]["url_public_api"]
		return url
	else
		identifier_match = match_figshare(dataname)
		if identifier_match isa RegexMatch
			identifier = identifier_match.match
			url = base_url(repo) * "/" *identifier
			return url
		else
			error("Please use a valid url, DOI, or Figshare ID")
		end
	end
end
```



Alternatively you could make those regular expressions just constants,
and then write ising `ismatch` and `match`.
In theory this is slower, but this is not performance critical,
and it tiny for a regex on such a short string.

```
function mainpage_url(repo::Figshare, dataname)
	if ismatch(doi_regex, dataname)
	    identifier = match(doi_regex, dataname).match
        down_url = base_url(repo) * "?doi=" * identifier
        doi_page = JSON.parse(text_only(getpage(down_url).root))
		url = doi_page[1]["url_public_api"]
		return url
	elseif ismatch(figshare_regex, dataname)
	    identifier = match(figshare_regex, dataname).match
		url = base_url(repo) * "/" *identifier
		return url
	else
		error("Please use a valid url, DOI, or Figshare ID")
	end
end
```


You could abstract this with a helper function, which might be useful elsewhere too
Though I am not sure that those big anon are args are very readabale.
I think this is over engineered.

```
function match_ifelse(pattern, str, matchedfunc, nomatchfunc)
	match_res = match(pattern, str)
	if match_res isa RegexMatch
		matchedfunc(match_res.match)
	else
		nomatchfunc()
	end
end


function mainpage_url(repo::Figshare, dataname)
	match_ifelse(doi_regex, dataname,
		function(identifier) # Matched DOI
			down_url = base_url(repo) * "?doi=" * identifier
			doi_page = JSON.parse(text_only(getpage(down_url).root))
			url = doi_page[1]["url_public_api"]
			return url
		end,
		function() #No match
			match_ifelse(figshare_regex, dataname,
				function(identifier) # Matched Figshare
					url = base_url(repo) * "/" *identifier
					return url
				end,
				function() # No match
					error("Please use a valid url, DOI, or Figshare ID")
				end
			)
		end
		)
	end
end
```



These are some options.
(I got a bit carried away).


Functions that sometimes return one argument and sometimes return  arguments,
are pretty unusual in julia,
and so are likely to catch future maintainers offguard.
The return type of `match` (`Union{RegexMatch, nothing}`) is a bit unusual, but less so and consistant with RegexMatch.

src/DataDepsGenerators.jl
oxinabox: Huh, I didn't think that threw an `ErrorException` I thought it threw something else.
Anyway for some alternatives see
 https://github.com/oxinabox/DataDepsGenerators.jl/pull/34/files#r197491537

oxinabox:
> My thoughts are: #30 in general gives amazing results but, have various attributes missing. However all of them mention the source of the data explicitly. What if we have individual types like Figshare() and Dataverse() and a general JSONLD() which eventually gets the JSON-LD, parses the source, and passes it into one of the types which we have already integrated.


I had a perhaps slightly more ambitions thought along those lines.
Or perhaps just a bit more silly.

Notion being to define a version of `generate`
that doesn't take a repo, and tries all of them (in parallel),
assesses the returned Metadata** of those that don't throw exceptions
for having the least missing fields and for having the important fields present (like download URL).
and selects the best.
Or maybe merges them.
Then generates the code.


Pseudocode follows:

```
function generate(dataname)
	failures = Vector()
	results = PriorityQueue()
	
	asyncmap(all_subtypes(DataRepo)) do repotype
		try
			metadata = find_metadata(repotype(), dataname)
			score = evaluate_metadata(metadata)
			results[score] = metadata
		catch err
			# We want to remember the failure,
			# so if we can't find any that work, we can explain what we tried,
			# and people can make good bug reports if one of those should have worked
			
			push!(failures, (repotype, err))
		end
	end
	
	if length(results) > 0
		# We got (at least) one
		metadata = first(results) # get the best
		generate_code(metadata)
	else
		# Got none
		println("We were unable to find usable metadata for dataname")
		println("We tried the following methods, and they failed for the reason listed")
		println()
		for (repo, err)
			println("# $repo")
			println(err)
			println()
		end
	end
end
```

I think it is more general to just try everything and see what works and how well it does.
Rather than rely on trying to use one API (which may or may not exist/be complete) to discover a better one.
By running them asynchronously it should be pretty fast, as it won't have to wait for pages to download.

I guess you could call this the "Throw everything at the wall and see what sticks" method of finding the metadata.

(**we'ld probably need to add more optional fields to the Metadata type. Like all the kinda standard stuff that is currently in the free text description)

src/Figshare.jl
```diff
@@ -0,0 +1,74 @@
...
+    urls
+end
+
+function get_checksums(repo::Figshare, page)
+    checksums = []
+    for i in page["files"]
+        push!(checksums, (:md5, i["computed_md5"]))
+    end
+    checksums
```
oxinabox: I think this is better written using a list comprehension.
`[(:md5, ii["computed_md5"]) for ii in page["files"]`

(I prefer `ii` over `i`, as a habit, as you can find it easily with simple search)

SebastinSanty:
Yes, your sounds more ambitious with the asynchronous part. Sounds interesting and I am up for it. 
> tries all of them (in parallel)

But even if I am trying to create them, wouldn't I require individual parts (like `DataCite`, `Figshare`) to be working. Wouldn't implementing all of them separately and then clubbing them together as you have noted be the better way to proceed?

src/DataDepsGenerators.jl
oxinabox: Cool, better.
A slightly cleaner way of writing this (so as not to be using exception handling)
is 
```
function match_doi(uri::String)
    identifier_match = match(r"\b(10[.][0-9]{4,}(?:[.][0-9]+)*\/(?:(?![\"&\'<>])\S)+)\b", uri)
    identifier_match === nothing ? nothing : identifier_match.match
end
```

Also I just realized `catch ErrorException`
doesn't do what you think it does.
Julia doesn't have type-guarded catch.
So `catch ErrorException` actually creates a local variable called `ErrorException` that is set to the exception 
value.

What you want to write for guarded exception handling is 

```
catch err
    if err isa ErrorException
         # handle
    else
         rethrow()
     end
end
```

oxinabox:
>  Wouldn't implementing all of them separately and then clubbing them together as you have noted be the better way to proceed?

Yes, that is what I was meaning.
Make them all separately and club together the ones that don't throw exceptions for a given identifier.

PR #40, opened 2018-06-30
[WIP] Implement Dataverse API 

SebastinSanty:


oxinabox:
It would be good if you could jot down some thoughts on the blockers your encountering,
or the things you are not understanding.
Or your understanding of the API so far

SebastinSanty:
I tried for a day more. The issues are:
1) What is an identifier in the Dataverse API? There is an `id` field, `identifier` field both of which don't work with the specified API atleast not for dataverse.harvard.edu. 
2) What I have done till now, is to get the details using `persistentId`. But I am able to get files and checksums details only.

PR #41, opened 2018-07-02
Use new ReferenceTests 0.3 and update GitHub

oxinabox:
ReferenceTests 0.3 incorporates DeepDiff.
Which will make checking this much easier.
Which also means using TestSetExtensions is not required (not that that was working right anyway).

I also changed GitHub's generation to limit the amount of text it includes and to not include the URLs in the references (which should make it less fragile),
I was doing this mostly to test that the diffs were working right,
and they are


SebastinSanty:
Why are we removing GitHub urls? Wouldn't DataDeps.jl require it for downloading?

oxinabox:
> Why are we removing GitHub urls? Wouldn't DataDeps.jl require it for downloading?

we are just removing them from the references. so that the tests don't break when ever a unrelated commit is made to the tested repos.
they are still generated, as is demonstrated by the integration tests

SebastinSanty:
Right, my bad. I failed to notice you were removing it from Reference tests.

SebastinSanty (approved):

PR #43, opened 2018-07-04
Using JSONLD to retrieve information

SebastinSanty:


oxinabox:
I am a fan of the `handle_key` method.

If you swap there argument order around, if has a really elegant way of writing it,
using the splatting arguments. (I think they say varadic function?)
(and follows the preferred order of collections before keys)

You can define it as

```
handle_key(json, key, otherkeys...) = get(json,  key) do  # do block is called when they is not found
     handle_key(json, otherkeys...) # This key missed, so try the others
end
handle_key(json) = nothing
```

Demo:
```
julia> foo = Dict("a"=>2)
Dict{String,Int64} with 1 entry:
  "a" => 2

julia> handle_key(foo, "a")  |> println
2

julia> handle_key(foo, "b") |> println
nothing

julia> handle_key(foo, "b", "a") |> println
2
```

SebastinSanty:
Right, I was thinking of adding `kwargs` like multiple arguments for `handle_keys`, which can check in the given order till the correct key is found.

oxinabox:
> which can check in the given order till the correct key is found.

That is what the code I posted does. It recursively peals off the first key until it is found, or until there is no keys left.
It just does so via dispatch and splatting

oxinabox:
We should probably check that the JSON-LD has the key-value pair
```
"@context": "http://schema.org",
```
And if it doesn't erroring out; or at least `warn`ing

I believe `@context` is JSON-LD for namespaces. Or something like that
So if it doesn't have that we can't know that it is using the names we expect.

On the other hand we could just not worry about it: Garbage-in => garbage out

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
+abstract type JSONLD <: DataRepo
+end
+
+export JSONLD_Web, JSONLD_DOI
```
oxinabox, line 4: All exports go in src/DataDepsGenerators.jl

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
...
+    Date: $(date)
+
+    $(desc)
+    """
+end
+
+function get_license(mainpage)
+    license = handle_keys("license", "", mainpage)
+    if license != nothing
+        if isa(license, String)
```
oxinabox: `isa` is available in infix notation 
`license isa String`

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
...
+            return license
+        elseif isa(license, Dict)
+            return handle_keys("url", "text", license)
+        end
+    end
+end
+
+function handle_keys(key1::String, key2::String, json)
```
oxinabox: I do want this changed to the have collection first then keys as a vararg.
Preference arguement order: https://github.com/JuliaLang/julia/issues/19150#issuecomment-320811138
Collection then Key.

Varargs:
https://docs.julialang.org/en/v0.6.1/manual/functions/#Varargs-Functions-1

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
...
+    if get(json, key1, nothing) == nothing
+        return get(json, key2, nothing)
+    else
+        return get(json, key1, nothing)
+    end
+end
+
+function get_urls(repo::JSONLD, page)
```
oxinabox: There is no need for both initializing this and for having an else statement.
Do one or the other.

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
...
+    url_list = handle_keys("distribution", "", page)
+    if url_list != nothing
+        urls = [handle_keys("contentUrl", "", ii) for ii in url_list if handle_keys("contentUrl", "", ii) != nothing]
+    else
+        urls = []
+    end
+    urls
+end
+
+function get_checksums(repo::JSONLD, page)
```
oxinabox: I think we should just define:
`get_checksums(::DataRepo) = nothing`
And then only overwrite it when we have something.
Rather than repeating it for all types.

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,86 @@
...
+function data_fullname(::JSONLD, mainpage)
+    mainpage["name"]
+end
+
+function website(::JSONLD, mainpage_url, mainpage)
+    mainpage_url
+end
+
+function mainpage_url(repo::JSONLD, dataname)
+    #We are making it work for both figshare id or doi
```
oxinabox: this comment doesn't belong here.

Also shouldn't this method be on `JSONLD_Web` ?

src/JSONLD/JSONLD_DOI.jl
```diff
@@ -0,0 +1,10 @@
+struct JSONLD_DOI <: JSONLD end
+
+function mainpage_url(repo::JSONLD_DOI, dataname)
+    if match_doi(dataname) != nothing
+        url = joinpath("https://data.datacite.org/", match_doi(dataname))
+        resp = HTTP.get(url, ["Accept"=>"application/vnd.schemaorg.ld+json"]; forwardheaders=true)
+        json = JSON.parse(resp.body |> String |> strip)
```
oxinabox: If you are going to use the pipe operator, use it completely, or not at all.
`resp.body |> String |> strip) |> JSON.parse`

Or

`JSON.parse(strip(String(resp.body)))`
Also, is the `strip` required?

src/JSONLD/JSONLD.jl
oxinabox: You have this method in `JSONLD_Web` already.
So delete this one.

test/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,9 @@
+using DataDepsGenerators
+using Base.Test
+
+using ReferenceTests
+
+@testset "JSONLD test" begin
+    @test_reference "../references/JSONLD_Web Kaggle.txt" generate(JSONLD_Web(), "https://zenodo.org/record/1287281")
```
oxinabox: The title says `Kaggle` but the URL says `zenodo`.
We want to test both Kaggle, and Zenodo,
and DataVerse, 
and FigShare via Web

test/references/JSONLD_Web Kaggle.txt
```diff
@@ -0,0 +1,20 @@
+register(DataDep(
+    "Map of Co-Seismic Landslides  for the M 7.8 Kaikoura, New Zealand Earthquake",
+    """
+    	Dataset: Map of Co-Seismic Landslides  for the M 7.8 Kaikoura, New Zealand Earthquake
+	Website: https://zenodo.org/record/1287281
+	Author: Valkaniotis Sotiris et al.
+	License: https://creativecommons.org/licenses/by/4.0/
+	Date: December 20, 2016
+	
+	<p>Prepared by the Research Group on Earthquake Geology in Greece (http://eqgeogr.weebly.com/)</p>
```
oxinabox: Looks like we need to strip any HTML out of the description.
We have Gumbo to parse HTML  and the `text_only` method already.
So should be easy to add a function in `utils.jl` for that

src/JSONLD/JSONLD.jl
SebastinSanty: This is how I used to do for abstracted types like `DataOnev2`.

oxinabox: well it isn't a big deal and it is trivial to change later. So I guess either is good.

src/utils.jl
```diff
@@ -18,6 +23,15 @@ text_only(doc::HTMLDocument) = text_only(doc.root)
 text_only(frag) = join([replace(text(leaf), "\r","") for leaf in Leaves(frag) if leaf isa HTMLText], " ")
 text_only(frags::Vector) = join(text_only.(frags), " ")
 
+function filter_html(random)
```
oxinabox: Split this into two methods.
`filter_html(::Void)=""`
and `filter_html(content) = ...`
Also `random` is a terrible name, `content` or `text` is better.

Splitting it is two avoids using `isa` in favor of dispatch,
which is more idiomatic julia

test/references/JSONLD_Web ICRISAT.txt
```diff
@@ -0,0 +1,15 @@
+register(DataDep(
+    "Phenotypic evaluation data of medium duration Pigeonpea advanced varieties trial",
+    """
+    	Dataset: Phenotypic evaluation data of medium duration Pigeonpea advanced varieties trial
+	Website: http://dataverse.icrisat.org/dataset.xhtml?persistentId=doi:10.21421/D2/ZS6XX1
+	Author: Sameer Kumar, CV, Anupama Hingane
+	License: <img src ="https://licensebuttons.net/l/by/4.0/88x31.png">
```
oxinabox: Looks like stripping the HTML is also required on this field.

And since license is potentially long,
it should probably be always last, after things like _date_

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,80 @@
...
+end
+
+function get_authors(repo::JSONLD, mainpage)
+    authors = handle_keys(mainpage, "author", "creator")
+    if authors isa Vector
+        return collect(skipmissing(handle_keys.(authors, "name")))
+    elseif authors isa Dict
+        return [handle_keys(authors, "name")]
+    else
+        return []
```
oxinabox: maybe add an `@assert(authors==nothing` 
if that is what you are expecting here.

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,80 @@
...
+    end
+    
+end
+
+function get_dates(repo::JSONLD, mainpage)
+    rawdate = handle_keys(mainpage, "datePublished", "dateCreated", "dateModified")
+    try
+        return Dates.format(Dates.DateTime(rawdate), "U d, yyyy")
+    catch error
+        if error isa MethodError
```
oxinabox: is this to handle `rawdate==nothing`
If so better to handle it if an `if rawdate==nothing` than with exception handling.

Exception handling can be the answer to this kind of thing I think.
But much more if there is a sequence of functions that might fail at anypoint,
and you want to treat them all the same.
For just one guarding against it with an `if` is better.
Julia (unlike say `python`) generally prefers avoiding exceptions to handling them.,

Also you forgot the `else rethrow()` in the catch block's `if`

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,80 @@
...
+    elseif license isa Dict
+        return handle_keys(license, "url", "text")
+    end
+end
+
+handle_keys(json, key, otherkeys...) = get(json,  key) do
+    handle_keys(json, otherkeys...)
+end
+
+handle_keys(json) = nothing
```
oxinabox, line 63: make this `missing`,
which will have to be done anyway later (when we move most things to be allowed to be missing and to properate missings everywhere)
and it will let `skipmissing` be used on URL's too.

A bunch of `nothing` conditions will need to become `missing` conditions, but that is ok and will need to be done anyway.

src/JSONLD/JSONLD.jl
```diff
@@ -0,0 +1,80 @@
...
+
+function get_urls(repo::JSONLD, page)
+    urls = []
+    url_list = handle_keys(page, "distribution")
+    if url_list != nothing
+        urls = [handle_keys(ii, "contentUrl") for ii in url_list if handle_keys(ii, "contentUrl") != nothing]
```
oxinabox: use `skipmissing` as discussed above

src/utils.jl
```diff
@@ -8,6 +8,11 @@ downloads and parses the page from the URL
 """
 getpage(url) = parsehtml(String(read(quiet_download(url))))
 
+# function parsehtml(junk::String)
```
oxinabox: Delete this commented out method?

src/utils.jl
```diff
@@ -18,6 +23,15 @@ text_only(doc::HTMLDocument) = text_only(doc.root)
 text_only(frag) = join([replace(text(leaf), "\r","") for leaf in Leaves(frag) if leaf isa HTMLText], " ")
 text_only(frags::Vector) = join(text_only.(frags), " ")
 
+function filter_html(random)
+    if random isa Void
+        return ""
+    end
+    if ismatch(r"<(\"[^\"]*\"|'[^']*'|[^'\">])*>", random)
```
oxinabox: I like this idea, that you check if it possibly could be HTML, using regex, before parsing it and stripping tags.
It makes for a bit faster then parsing unnecessarily.
It would be good to add a comment to this line saying: "Check if might be HTML"
so it is clear that is what is going on.

src/utils.jl
SebastinSanty: Yes, I was debugging and forgot to update 😅

src/JSONLD/JSONLD.jl
SebastinSanty: This is to handle the difference in the incoming date formats. Some have just the year and some have different formats of date.

src/JSONLD/JSONLD.jl
SebastinSanty: Didn't quite understand this

SebastinSanty:
Kaggle for some reason is down. Will restart the tests once that works.

src/JSONLD/JSONLD.jl
oxinabox: Under what circumstances does the else trigger?

src/JSONLD/JSONLD.jl
oxinabox: It is not good to catch all errors.
And when not clear (like this is to me) it is good to comment with what kind of thing us causing the error.

src/JSONLD/JSONLD.jl
SebastinSanty: How will I handle `if url_list != nothing` if I am using `missing`?

oxinabox: I'll make a PR on this and show you.
This is mergable right now I think

SebastinSanty: Sure, it will be helpful if you can.

oxinabox (approved):
very cool.
I think we have technically cracked the 150 million "datasets" mark with this.
I say technically because 97 million of them would be from CrossRef DOIs, which are not data as most people would consider them

PR #46, opened 2018-07-10
[WIP] Overhaul Metadata for individual components

SebastinSanty:
Please check and let me know what you think. I'll change across other Datarepos ones I get your reviews. Obviously the tests will fail.


oxinabox:
I am excited, I'll get reviewing this right away

src/DataDepsGenerators.jl
```diff
@@ -12,26 +12,60 @@ abstract type DataRepo end
 
 const Opt{T} = Union{Missing, T}
 struct Metadata
+    shortname::Opt{String}
     fullname::Opt{String}
     website::Opt{String}
     description::Opt{String}
+    author::Opt{String}
```
oxinabox: Lets keep all the authors.
So this can be an `Opt{Vector{String}}`
and then we can move `format_authors` out of each individual type.
Cutting them down to just converting dicts etc into strings

src/DataCite.jl
```diff
@@ -3,35 +3,27 @@ end
...
-function description(repo::DataCite, mainpage)
-    attributes = mainpage["attributes"]
-    desc = attributes["description"]
-    authors = join.([[names[2] for names in value] for value in attributes["author"]], " ")
-    author = format_authors(authors)
-    license = attributes["license"]
-    date = attributes["published"]
-    dataset_cite = citation_text(mainpage["id"])
-    
+description(repo::DataCite, mainpage) = (mainpage["attributes"]["description"] != nothing? mainpage["attributes"]["description"] : missing)
```
oxinabox: How was this coming up as `nothing` in the first place?
I would have expected emptystring, or a `KeyError`, but not `nothing`

I expect nothing from regex, and KeyErrors from JSON/Dicts

SebastinSanty: Yes, even I was wondering the same. Somehow the tests didn't pass and when I checked back on the old description content, there was this nothing condition there. I didn't investigate much.

src/DataCite.jl
```diff
@@ -3,35 +3,27 @@ end
...
+end
+
+license(::DataCite, mainpage) = mainpage["attributes"]["license"]
+
+publishedDate(::DataCite, mainpage) = mainpage["attributes"]["published"]
+
+datasetCite(::DataCite, mainpage) = citation_text(mainpage["id"])
+
+function paperCite(::DataCite, mainpage)
     paper_cite = nothing
```
oxinabox: this should be `missing` rather than  `nothing`

src/DataDepsGenerators.jl
```diff
@@ -46,41 +80,51 @@ include("Figshare.jl")
...
-    escape_multiline_string("""
-    Dataset: $(meta.fullname)
-    Website: $(meta.website)
-    $(meta.description)
-    """) |> strip
+    netString =  """
+    register(DataDep(
+        \"$(meta.shortname)\",
+        \"\"\""""
+    netString = format_meta(meta.website, netString, "Website")
```
oxinabox: change these to:

```
netstring *= format_meta(meta.website, "Website")
```

and make `format_meta` return the string to be appended.
Or an empty string, if nothing is to be appended.

etc.
Then you don't need to keep passing in the current string as an argument.

src/DataDepsGenerators.jl
```diff
@@ -46,41 +80,51 @@ include("Figshare.jl")
...
+    if ismissing(data)
+        return netString
+    else
+        new_data = (indentbool? indent(string(data)) : string(data))
+        if ismissing(label)
+            return netString * "\n" * new_data
+        else
+            return netString * "\n" * label * ": " * new_data
+        end
+    end
```
oxinabox: I suggest instead:

Make `indentbool` a kwarg called `indent_field`,
and default it to `true` since that is what you mostly have it set to.

and I suggest the default value for `label`  should be `""`  because it isn't a missing label, it is an empty label.

and use dispatch to handle the missings.

So it would look something like this:

```
format_meta(::Missing, args...; kwargs...) = ""

function format_meta(data::Any, label=""; indent_field=true)
     if label != ""
         label*=":"
     end
     field = (indent_field ? indent(string(data)) : string(data))
     return "\n" * label * field
end
```

oxinabox: and for example

```
netString = format_meta(meta.description, netString, missing,  true)
```

would become

```
netString *= format_meta(meta.description)
```

oxinabox: In general I'm not too sure on what `indentbool` is for,

Maybe it can be simplified to not having it at all.

eg.
```
format_meta(::Missing, args...; kwargs...) = ""

function format_meta(data::Any, label="")
     if label != ""
         label*=":\t"
     end
     field = string(data)
     return "\n" * label * field
end
```

Would do the same thing, I think.

I do like the idea of putting a tab between fieldnames and there values.
It seems semantic

src/DataDepsGenerators.jl
```diff
@@ -12,26 +12,60 @@ abstract type DataRepo end
...
 end
 
-function find_metadata(repo, dataname)
+function find_metadata(repo, dataname, shortname)
 
     mainpage, url = mainpage_url(repo, dataname)
+    fullname = data_fullname(repo, mainpage)
+    if shortname == nothing
+        shortname = reduce((s,r)->replace(s, r, ""), fullname, ['\\', '/', ':', '*', '?', '<', '>', '|'])
+    end
```
oxinabox: I think it is cleaner to have this in a function.

```
data_shortname(repo, shortname, fullname) = shortname
function data_shortname(repo, shortname::Void, fullname)
    # Remove any characters not allowed in a file path
    reduce((s,r)->replace(s, r, ""), fullname, ['\\', '/', ':', '*', '?', '<', '>', '|'])
end
```

src/DataDepsGenerators.jl
```diff
@@ -97,9 +141,9 @@ function format_checksums(csum::AbstractString)
     if length(csum)>0 "\"$csum\"" else "" end
 end
 
-function format_checksums(::Void)
-    ""
-end
+format_checksums(::Void) = missing
```
oxinabox: delete `format_checksums(::Void)`,
becomes things that are currently returning `nothing` for checksums should be returning `missing`.
Deleting this will help us find them

src/DataDepsGenerators.jl
```diff
@@ -110,7 +154,7 @@ function format_authors(authors::Vector)
         authors[1] * " et al."
     else
         warn("Not able to retrieve any authors")
-        "Unknown Author"
+        missing
```
oxinabox: I think you were right be for,
since this is being use to typeset authors,
if no authors are given to it the correct thing is not to typeset nothing,
but to typeset `"Unknown Authors"`, or `"Authors not specified"`
(The latter would match well with "License not specified")

I think also needed is 
`format_authors(authors::Missing) = "Authors not specified"`
So maybe for D.R.Y purposes,
specifying this as:
```
else
    warn ...
    format_authors(missing)
```

test/references/CKAN Consumer Complaint.txt
```diff
@@ -1,15 +1,13 @@
...
-    	Dataset: Consumer Complaint Database
-	Website: https://catalog.data.gov/dataset/consumer-complaint-database
-	Author:
-	License: License not specified
-	Date: February 26, 2014
-	Maintainer: devops@cfpb.gov
-	
-	These are complaints we’ve received about financial products and services.
-    """,
-    Any["https://data.consumerfinance.gov/api/views/s6ew-h6mp/rows.csv?accessType=DOWNLOAD", "https://data.consumerfinance.gov/api/views/s6ew-h6mp/rows.csv?accessType=DOWNLOAD", "http://data.consumerfinance.gov/api/views.json", "http://data.consumerfinance.gov/api/views.xml", "http://data.consumerfinance.gov/api/views/s6ew-h6mp.json"],
```
oxinabox: Looks like you for got tin include the line that adds the dataset's full-name to the top.

src/DataDepsGenerators.jl
```diff
@@ -12,26 +12,60 @@ abstract type DataRepo end
...
+author(::DataRepo, mainpage) = missing
+maintainer(::DataRepo, mainpage) = missing
+license(::DataRepo, mainpage) = missing
+publishedDate(::DataRepo, mainpage) = missing
+createDate(::DataRepo, mainpage) = missing
+modifiedDate(::DataRepo, mainpage) = missing
+paperCite(::DataRepo, mainpage) = missing
+datasetCite(::DataRepo, mainpage) = missing
+get_checksums(::DataRepo, mainpage) = missing
+get_urls(::DataRepo, mainpage) = missing
```
oxinabox: Good stuff.
esp that the website has different (and sane) default

test/references/DataCite Ecology.txt
```diff
@@ -1,20 +1,12 @@
 register(DataDep(
-    "Data from Ecology and genomics of an important crop wild relative as a prelude to agricultural innovation",
+    "Data from: Ecology and genomics of an important crop wild relative as a prelude to agricultural innovation",
```
oxinabox: That is not right.
A DataDep name can not contain the `:` character.
As it has to be a valid filepath.
Something has gone wrong somewhere with the `shortname` stuff

test/references/DataCite Ecology.txt
```diff
@@ -1,20 +1,12 @@
...
-	if you use this in your research.
-    """,
-    String["PUT DOWNLOAD URL HERE"],
-    
-))
+Website: https://doi.org/10.5061/dryad.74699
+Author: 	Eric J. B. Von Wettberg et al.
+License: 	https://creativecommons.org/publicdomain/zero/1.0/
+Date of Publication: 	2018
+Please cite this paper: 	Von Wettberg, E. J. B., Chang, P. L., Başdemir, F., Carrasquila-Garcia, N., Korbu, L. B., Moenga, S. M., … Cook, D. R. (2018). Ecology and genomics of an important crop wild relative as a prelude to agricultural innovation. Nature Communications, 9(1). doi:10.1038/s41467-018-02867-z
```
oxinabox: I think the label for `papercite` and `datacite`  should begin with an extra newline.
i.e. "\nPlease cite this paper:"

src/DataDepsGenerators.jl
```diff
@@ -46,41 +80,51 @@ include("Figshare.jl")
...
+        \"$(meta.shortname)\",
+        \"\"\""""
+    netString = format_meta(meta.website, netString, "Website")
+    netString = format_meta(format_authors(meta.author), netString, "Author", true)
+    netString = format_meta(meta.maintainer, netString, "Maintainer", true)
+    netString = format_meta(meta.license, netString, "License", true)
+    netString = format_meta(meta.publishedDate, netString, "Date of Publication", true)
+    netString = format_meta(meta.createDate, netString, "Date of Creation", true)
+    netString = format_meta(meta.modifiedDate, netString, "Date of Modification", true)
+    netString = format_meta(meta.description, netString, missing,  true)
```
oxinabox: add a `netString *= "\n"`
before `format_meta(meta.description)

to leave a gap between the "Front matter" of mostly short fields,
and the description/citation details

src/DataDepsGenerators.jl
```diff
@@ -46,41 +80,51 @@ include("Figshare.jl")
...
-    $(meta.description)
-    """) |> strip
+    netString =  """
+    register(DataDep(
+        \"$(meta.shortname)\",
+        \"\"\""""
+    netString = format_meta(meta.website, netString, "Website")
+    netString = format_meta(format_authors(meta.author), netString, "Author", true)
+    netString = format_meta(meta.maintainer, netString, "Maintainer", true)
+    netString = format_meta(meta.license, netString, "License", true)
```
oxinabox: license should be after all the date fields,
as sometimes it is very long.
So it should be the last of the "frontmatter" short fields.

src/DataDepsGenerators.jl
```diff
@@ -46,41 +80,51 @@ include("Figshare.jl")
...
+    netString = format_meta(meta.publishedDate, netString, "Date of Publication", true)
+    netString = format_meta(meta.createDate, netString, "Date of Creation", true)
+    netString = format_meta(meta.modifiedDate, netString, "Date of Modification", true)
```
oxinabox: "Date of **Last** Modification:

src/DataCite.jl
oxinabox: It is worth investigating further, 
though free free to create an issue on the repo and mark it as something to come back to.
You might have to ask someone in #data who knows about JSON.jl if there is a reason this might occur.

src/CKAN.jl
```diff
@@ -1,23 +1,20 @@
...
-    
-    """
-    Author: $(author)
-    License: $(license)
-    Date: $(date)
-    Maintainer: $(maintainer)
-
-    $(desc)
-    """
+    Dates.format(rawdate, "U d, yyyy")
```
oxinabox, line 16: Do you thing we could do more date processing at at the metadata level?
We could allow  the Metadata to contain `Union{Missing, Dates.Date,  String}`
where the String is often just a year or some such.

then define:
```
format_dates(rawdate) = rawdate#Catchs the missing and the string
format_dates(rawdate::Dates.Date) = Dates.format(rawdate, "U d, yyyy")
```

Which would I think eliminate this line from most of the generators.
More D.R.Y

And it would later easily let us change the format for everything, since it would be in just one place

oxinabox:
I am finished my reviews
Looking really good.
I think this will be a big improvement for maintainability.

I also like that you are using more single line functions
(when the lines are short enough)

src/DataCite.jl
SebastinSanty: https://api.datacite.org/works/10.5061/dryad.74699 `description` is `null`

oxinabox: Ah right, JSON being a subset of javascript has an actual `null`.
In 0.7 this code can be written `something(mainpage["attributes"]["description"], missing)`
For now this will do, or `Compat` could be added as a dependency to give that function.

src/CKAN.jl
```diff
@@ -26,7 +25,7 @@ function get_urls(repo::CKAN, page)
 end
 
 function get_checksums(repo::CKAN, page)
```
oxinabox: You can delete this outright, and let it fall back to the default

src/DataCite.jl
```diff
@@ -33,7 +31,7 @@ function get_urls(repo::DataCite, page)
 end
 
 function get_checksums(repo::DataCite, page)
```
oxinabox: You can delete this outright too, and let it fall back to the default

src/DataDepsGenerators.jl
```diff
@@ -19,9 +19,9 @@ struct Metadata
     author::Opt{String}
     maintainer::Opt{String}
     license::Opt{String}
-    publishedDate::Opt{String}
-    createDate::Opt{String}
-    modifiedDate::Opt{String}
+    publishedDate::Union{DateTime, Opt{String}}
+    createDate::Union{DateTime, Opt{String}}
+    modifiedDate::Union{DateTime, Opt{String}}
```
oxinabox: It does the same thing, but I think more semantic is
`Opt{Union{DateTime,String}}`

src/DataDepsGenerators.jl
```diff
@@ -32,16 +32,14 @@ function find_metadata(repo, dataname, shortname)
...
-    end
+    shortname = data_shortname(repo, shortname, fullname)
 
     Metadata(
         fullname,
         shortname,
         website(repo, url, mainpage),
         description(repo, mainpage),
-        author(repo, mainpage),
+        format_authors(author(repo, mainpage)),
```
oxinabox: MetaData  should have a field `authors::Opt{Vector{String}}`.
So all the authors can be stored.
And the `format_authors` belong with the other formatting commands (like `format_date`),
in the part that generates the code

src/DataDepsGenerators.jl
```diff
@@ -84,38 +92,39 @@ function message(meta)
...
-    netString = format_meta(meta.description, netString, missing,  true)
-    netString = format_meta(meta.paperCite, netString, "Please cite this paper", true)
-    netString = format_meta(meta.datasetCite, netString, "Please cite this dataset", true)
-    netString = strip(netString * "\n\"\"\",")
-    netString = format_meta(meta.dataurls, netString)
-    netString = netString * ","
-    netString = format_meta(format_checksums(meta.datachecksums), netString)
-    netString = strip(netString * "\n))")
+    netString *= format_meta(meta.website, "Website")
+    netString *= format_meta(meta.author, "Author")
```
oxinabox: see this should be:
`format_meta(format_authors(meta.authors), "Author")`

src/DataDepsGenerators.jl
```diff
@@ -84,38 +92,39 @@ function message(meta)
...
+
+function format_meta(data::Any, label=""; indent_field=true)
+    if label != ""
+        label*=":"
     end
+    field = (indent_field ? indent(string(data)) : string(data))
+    return "\n" * label * field
 end
 
+format_authors(authors::Missing) = "Authors not specified"
```
oxinabox: This should be down next to the other `format_authors`

src/DataDepsGenerators.jl
SebastinSanty: Then I have to change the metadata struct type. I have used format_authors when instantiating a new metadata in `find_metadata()`.

src/DataDepsGenerators.jl
```diff
@@ -158,6 +167,10 @@ function format_authors(authors::Vector)
     end
 end
 
+function handle_null(attr::Any)
+    attr != nothing? attr : missing
+end
```
oxinabox: This belongs in `utils.jl` next to `lift`
And it should be written using dispatch on the type of its input.
You know what I mean?

src/DataDryad.jl
```diff
@@ -3,35 +3,30 @@ end
...
-    dateelem = matchall(sel".publication-header p", mainpage.root)
-    date = replace(text_only(dateelem[length(dateelem)-1]), "Date Published: ", "")
-    paper = text_only(first(matchall(sel".citation-sample", mainpage.root)))
-    dataset = replace(text_only(last(matchall(sel".publication-header p", mainpage.root))), "DOI: ", "")
-
-    """
-    Author: $(author) et. al.
-    License: $(license)
-    Date: $(date)
+    println(author)
```
oxinabox: I guess this was for debugging and you forgot to delete it

SebastinSanty:
Yep

test/references/CKAN Consumer Complaint.txt
```diff
@@ -1,12 +1,12 @@
...
     "Consumer Complaint Database",
     """
-    
-Website: https://catalog.data.gov/dataset/consumer-complaint-database
-Author: 	
-Maintainer: 	devops@cfpb.gov
-License: 	License not specified
-Date of Creation: 	February 26, 2014
+Website:	https://catalog.data.gov/dataset/consumer-complaint-database
+Author:
```
oxinabox: Looks like something is going wrong with the Author here.
I am guessing it is an empty, string
we might need a dispatch that catches that, and re-dispatches it as a missing

test/references/CKAN Gold Prices.txt
```diff
@@ -1,12 +1,12 @@
...
     "Gold Prices in London 1950-2008 (Monthly)",
     """
-    
-Website: https://demo.ckan.org/dataset/gold-prices
-Author: 	Bundesbank
-Maintainer: 	Rufus Pollock
-License: 	Open Data Commons Public Domain Dedication and License (PDDL)
-Date of Creation: 	July 18, 2017
+Website:	https://demo.ckan.org/dataset/gold-prices
+Author:	Bundesbank
```
oxinabox: I just notices.
Why don't these ones get tabs?
Tabs or no tabs I don't mind but consistency is good

test/references/DataCite Ceramic.txt
```diff
@@ -1,19 +1,14 @@
 register(DataDep(
     "Study Of Surface Roughness With The Variation In Applied Load Of Rutile Ceramic Reinforced Aluminium Composite",
     """
-    	Dataset: Study Of Surface Roughness With The Variation In Applied Load Of Rutile Ceramic Reinforced Aluminium Composite
-	Website: https://doi.org/10.5281/zenodo.1147572
-	Author: Rama Arora
-	License: https://creativecommons.org/licenses/by/4.0/
-	Date: 2018
-	
+Website:	https://doi.org/10.5281/zenodo.1147572
```
oxinabox: Still missing dataset name,

src/DataDepsGenerators.jl
```diff
@@ -54,6 +52,16 @@ function find_metadata(repo, dataname, shortname)
     )
 end
 
+function data_shortname(repo, shortname::Void, fullname)
```
oxinabox, line 55: Colons `:` are still showing up in the names. Which is not legal.
Also unescaped `"`.
`data_shortname` should strip both
Not sure why it isn't.

src/DataDepsGenerators.jl
oxinabox: Maybe it should be called `miss_null`
or `nothing2missing` or something.

handle is very vague as a word

oxinabox:
Getting there.

As some point some things became 
camelCased

Like
 - publishedDate
 - createDate
 - modifiedDate
 - paperCite
 - datasetCite

Can they be changed to snake_case

SebastinSanty:
Sure, I'll switch. Personally I prefer camelcase that's why it came in.

oxinabox:
> Sure, I'll switch. Personally I prefer camelcase that's why it came in.

I've stopped having personal preference for casing;
since I feel consistency within project and with other projects matters more.
(I even give up on my more strongly held preference for tabs over spaces :-o)

src/DataDepsGenerators.jl
oxinabox: yes indeed.

src/DataOneV2/KnowledgeNetworkforBiocomplexity.jl
```diff
@@ -2,5 +2,5 @@ struct KnowledgeNetworkforBiocomplexity <: KNB
 end
 
 function pub_date(repo::KnowledgeNetworkforBiocomplexity, mainpage)
-    "", "No dates released"
+    "No dates released"
```
oxinabox, line 5: This can just be deleted to fall back to default `missing`

oxinabox:
I have noticed the indenting of the final output is all screwed up.
It needs to be:
```
regisister(DataDep(
    "NAME",
    """
    MSG1
    MSG2
    """,
   ["URL1", "URL2]
    "CHECKSUM"
))
```

Every line on the inside needs to be indented.
At the start of each line.

Take a look at the string at the end of the current current structure:
https://github.com/oxinabox/DataDepsGenerators.jl/blob/b2e5bc460930dcde4656ab750487b5188e8741f0/src/DataDepsGenerators.jl#L62-L81
and note that it does include indents

Not that the `indent` function exits for indenting multiline strings.
It splits the string into each line and then appends a tab to each, before joining it back into one string.
So it was handling the multiline string `message` ware returning.

Related:
I think that `message` function should be only generating the free-text message (which is the name of that field in the DataDep structure).
Rather than the whole thing.


---
I think this is all generally on the right track,
can you go through every concern I raised (you'll have to click "expand outdated in github"),
and I will do a code review once everything is done.

SebastinSanty:
Pinging on the PR

oxinabox:
The Dataset fullname is still missing from the message text.
It is need there.l
Other than that looks mostly good.
I've not checked over my full list of issues I raised with earlier ones.
So hopefully you have.

I am pretty hugely busy today, but I will merge it once that missing dataset name in messages is fixed, assuming you are happy with it.

Other changes can come in other PRs.

oxinabox:
🎉

PR #47, opened 2018-07-19
Multisource data acquisition

SebastinSanty:


SebastinSanty:
@oxinabox I am not sure, but the datatypes are not matching for some reason. Can you have a look?

PS: I am running only `include("test/CKAN.jl")` and only one generator for now without any async part, so as to debug and localize errors easily.

Also, your reviews on the implementation will be really helpful.

SebastinSanty:
For your quick reference, this is the datatypes I am returning:
```
String
String
String
String
Array{AbstractString,1}
String
String
String
String
String
Array{Any,1}
String
```
And these are the metadata datatypes:
```
shortname::Opt{AbstractString}
fullname::Opt{AbstractString}
website::Opt{AbstractString}
description::Opt{AbstractString}
author::Opt{Vector{AbstractString}}
maintainer::Opt{AbstractString}
license::Opt{AbstractString}
published_date::Opt{Union{Date, DateTime,AbstractString}}
create_date::Opt{Union{Date, DateTime,AbstractString}}
modified_date::Opt{Union{Date, DateTime,AbstractString}}
paper_cite::Opt{AbstractString}
dataset_cite::Opt{AbstractString}
dataurls::Opt{Vector}
datachecksums::Any
```

SebastinSanty:
Hi @oxinabox , I think this missed to hit your mailbox. A small review will be appreciated :)

oxinabox:
This is not great.

You need to make sure your code  is consistent and logical for how things of different types are combined.
This means that merging 2 `missing`s should give a `missing`, not an empty string.
Similar for what happens with merging vectors.
This is why you are getting type errors.

In general you really want to be constructing 
a function that takes in
something like
`combine(afield, bfield, Val{fieldname}())`
Which you could then invoke in a loop (or via `reduce`).
You only need pairwise rules for combining each field.

That way you can write dispatches, like
`combine(::Missing, bfield, ::Val)=bfield`,
or
```
function combine(aauthors::Vector, bauthors::Vector, ::Val{:authors})
	if length(aauthors) > length(bauthors)
		aauthors
	else 
		bauthors
	end
end
```
Or a rule that says the shorter (nonzero) license text is better.
Etc.
To express the rules concisely and clearly.
You want to consider how the code should be structured?



Global variables are also bad.
Which I feel you should be well aware of.
They are not being useful here at all;
and are a serious code-smell.


I think maybe you want to take a few steps back, and take another swing at this PR.
With a fresh branch from master.
Think about what exactly the goals are;
in a precise way.
And then break it down a bit more;
before you start coding.
You could start you new PR,
just with the changes to the Readme that will be required to explain to the user how to use the new feature you'll be implementing.
Then in that PR you can outline your plans for what to change, and how,
before you code it.
Then I can give feedback on that.

SebastinSanty:
I had made some very silly mistakes. Anyways, I have solved them. Tests are passing now. I have accounted to what you have said above. But I wasn't sure of how we can work with pairwise combining as that will lead to O(n^2), which can be easily be avoided. Please let me know your thoughts on the current version. Also the `create_generators()` will create Array of generators suitable for the purpose. Eg: If it is a DOI it makes sense to work with certain generators like DataCite, Figshare. There is no need to scan through all of the generators.

I'll continue this for multiple generators (`try catch` for different fields in generators is in a different branch), and the async part after your feedback.

oxinabox:
Consider this:
While it is pairwise, the result is made by comparing (combinging) the current result to the next in a sequence/

This the functional `reduce` (or `fold`) operation.
(You can look it up on wikipedia or similar no doubt. Or in the julia docs)



```
julia> combine(::Missing, ::Missing) = missing
combine (generic function with 1 methods)

julia> combine(::Missing, x) = x
combine (generic function with 2 methods)

julia> combine(x, ::Missing) = x
combine (generic function with 3 methods)

julia> combine(x,y) = length(x) > length(y) ? x : y
combine (generic function with 4 methods)

julia> function combine_all(values)
               ret = missing
               for value in values
                       print("merging $ret \t with $value")
                       ret = combine(ret, value)
                       println("\t\t-------------> ", ret)
               end
               ret
       end
combine_all (generic function with 1 method)

julia> combine_all([missing, missing, "abcde", missing, "ab", "longest", "short"])
merging missing          with missing           -------------> missing
merging missing          with missing           -------------> missing
merging missing          with abcde             -------------> abcde
merging abcde            with missing           -------------> abcde
merging abcde            with ab                -------------> abcde
merging abcde            with longest           -------------> longest
merging longest          with short             -------------> longest
"longest"
```

oxinabox:
Re: 
`remove_cite_version`
That needs some comments explaining why,
and I kind of feel we shouldn't do this -- that information is useful 

I am guessing it is because the same generator, on different days, includes that information or doesn't include that information (I feel like I've seen that happen before).
So not removing it breaks are test.
If that is the case,
then we should raise a bug upstream, that they are not returning consistent results.
and we should do a workaround in the **testing script only**, to strip that string.
Like how the github tests strips out all URLS

oxinabox:
re: ` rule_dispatcher`
I think you are thinking on the right lines but take a good look at what documentation and Stackoverflow posts, you can find on "Dispatch On Value" in julia, -- that is things like `Val{:authors)`,
and maybe play around with solving a toy problem using it,
to get a good feel for it as a tool to solve problems.
It will make the code a fair bit cleaner.

oxinabox:
> Also the create_generators() will create Array of generators suitable for the purpose. Eg: If it is a DOI it makes sense to work with certain generators like DataCite, Figshare. There is no need to scan through all of the generators.

But why not scan through all the generators?

For some generators (e.g. JSON-LD) it is impossible to know if it will work without downloading the page.
At which point it will throw an exception.

For others like all the ones that only eat DOIs, it will throw an exception if it doesn't `match_doi`,
in which case it is basically takes the same amount of time as  `create_generators()`.

And we should be be able to use Async (or if it comes to it outright multiprocessing parallelism),
to make checking extras take approximately the same wall-time as checking just one.

src/DataCite.jl
```diff
@@ -24,11 +24,11 @@ function paper_cite(::DataCite, mainpage)
...
-function get_urls(repo::DataCite, page)
-    urls = ["PUT DOWNLOAD URL HERE"]
-    info("DataCite based generation can only generate partial registration blocks, as DataCite metadata does not (currently) include the URL to the resource. You will have to edit in the URL after generation.")
-    urls
-end
+# function get_urls(repo::DataCite, page)
+#     urls = ["PUT DOWNLOAD URL HERE"]
+#     info("DataCite based generation can only generate partial registration blocks, as DataCite metadata does not (currently) include the URL to the resource. You will have to edit in the URL after generation.")
+#     urls
+# end
```
oxinabox: Why is this commented out?

src/DataDepsGenerators.jl
```diff
@@ -190,4 +236,8 @@ function mainpage_url(repo::DataRepo, dataname)
     getpage(url), url
 end
 
+function remove_cite_version(code)
+    replace(code, r"\(Version 1\) " => "")
+end
```
oxinabox: Move this function into `utils.jl` 
and say `replace(code, r"\(Version \d+\) " => "")`
to get any number of numbers

src/DataDepsGenerators.jl
```diff
@@ -87,6 +87,48 @@ include("DataCite.jl")
 include("Figshare.jl")
 include("JSONLD/JSONLD.jl")
 
+function aggregate(generator_meta)
```
oxinabox: `generator_meta` is not a good variable name.
the importance of `generator` is not clear.
and it is not a plural.
variables that contain lists, should be plural; or a least names such that they are obviously not just a single value.


Maybe you want `metadatas`

src/DataDepsGenerators.jl
```diff
@@ -87,6 +87,48 @@ include("DataCite.jl")
...
+end
+
+function combine_all(values::Vector, sym::Symbol)
+    ret = missing
+    for value in getfieldlist(values, sym)
+        ret = combine(ret, value, sym)
+    end
+    ret
+end
+
```
oxinabox, line 121: You don't quiet have dispatch on value working correctly.
Talk to me, or someone on slack.

src/DataDepsGenerators.jl
```diff
@@ -125,14 +167,18 @@ function format_meta(data::Any, label="")
...
-    meta = find_metadata(repo, dataname, shortname)
-
-    body(meta)
+    generators = []
+    generator_meta = []
+    push!(generators, repo)
+    for ii in generators
+        push!(generator_meta, find_metadata(ii, dataname, shortname))
+    end
+    aggregate(generator_meta)
```
oxinabox: Stick into  `utils.jl`
```
"""
    leaf_subtypes(T)
Returns all the nonabstract types decedent from `T`.
"""
function leaf_subtypes(T)
       if isleaftype(T)
           T
       else
           vcat(leaf_subtypes.(subtypes(T))...)
       end
end
```  

Into the main file for generation stick something like

```
function generate(repo::DataRepo, dataname; kwargs...)
    generate([repo], dataname; kwargs...)
end

function generate(dataname; kwargs...)
    all_repos = [T() for T in leaf_subtypes(DataRepo)]
    generate(all_repos , dataname; kwargs...)
end


function generate(repos, dataname, shortname = nothing, show_failures=false)
	retrieved_metadatas = MetaData[]
	failures = Tuple{DataRepo, Exception}[]
	
	# Get all the metadata we can
	@sync for repo in repos
		@async try
			metadata = find_metadata(repo, dataname, shortname)
			push!(retrieved_metadatas, metadata)
		catch err
			push!(failures, (repo, err))
		end
	end
	
	# Displace errors if required
	if length(retrieved_metadatas) == 0 || show_failures
		for (repo, err) in failures
			println(repo, " failed due to")
			println(err)
			println()
		end
	end
	
	# merge them
	aggregate(retrieved_metadatas)
end
```

I haven't tested this code.
But hopefully you can get the idea.

test/JSONLD/JSONLD.jl
```diff
@@ -4,10 +4,10 @@ using Base.Test
...
-    @test_reference "../references/JSONLD_Web Zenodo.txt" generate(JSONLD_Web(), "https://zenodo.org/record/1287281")
-    @test_reference "../references/JSONLD_Web ICRISAT.txt" generate(JSONLD_Web(), "http://dataverse.icrisat.org/dataset.xhtml?persistentId=doi:10.21421/D2/ZS6XX1")
-    @test_reference "../references/JSONLD_Web Figshare.txt" generate(JSONLD_Web(), "https://figshare.com/articles/_shows_examples_of_coordinated_and_uncoordinated_motion_for_dangerous_and_non_dangerous_crowd_behavior_/186003")
+    # @test_reference "../references/JSONLD_Web Kaggle.txt" generate(JSONLD_Web(), "https://www.kaggle.com/stackoverflow/stack-overflow-2018-developer-survey")
+    # @test_reference "../references/JSONLD_Web Zenodo.txt" generate(JSONLD_Web(), "https://zenodo.org/record/1287281")
+    # @test_reference "../references/JSONLD_Web ICRISAT.txt" generate(JSONLD_Web(), "http://dataverse.icrisat.org/dataset.xhtml?persistentId=doi:10.21421/D2/ZS6XX1")
+    # @test_reference "../references/JSONLD_Web Figshare.txt" generate(JSONLD_Web(), "https://figshare.com/articles/_shows_examples_of_coordinated_and_uncoordinated_motion_for_dangerous_and_non_dangerous_crowd_behavior_/186003")
```
oxinabox: these have to be uncommented out

oxinabox (changes requested):
Getting there.
In one of the comments I wrote the code for how to actually make this async.

Other than that the main issue remaining is that your disaptch on value doesn't actually do any dispatch on value

src/DataDepsGenerators.jl
oxinabox: See:

https://stackoverflow.com/questions/51568006/how-do-i-do-dispatch-on-value-in-julia#51568007

test/JSONLD/JSONLD.jl
SebastinSanty: Issues with incoming JSONLD. Not able to JSON parse them.

src/DataCite.jl
SebastinSanty: Because, `"PUT DOWNLOAD URL HERE"` is not a valid url and when comparing the length, a valid url array with length 1 will get ignored if this goes in. This can probably shift to the final makeup of the registration block, in which `missing` data_urls will be replaced with the `"PUT DOWNLOAD URL HERE"`.

oxinabox: Cool makes sense.
I think even just leaving the URL as missing is fine.
Delete this rather than just commenting it out.

oxinabox:
Reposting this in the main thread so it doesn't get vanished by github:
----------------

Stick into  `utils.jl`
```
"""
    leaf_subtypes(T)
Returns all the nonabstract types decedent from `T`.
"""
function leaf_subtypes(T)
       if isleaftype(T)
           T
       else
           vcat(leaf_subtypes.(subtypes(T))...)
       end
end
```  

Into the main file for generation stick something like

```
function generate(repo::DataRepo, dataname; kwargs...)
    generate([repo], dataname; kwargs...)
end

function generate(dataname; kwargs...)
    all_repos = [T() for T in leaf_subtypes(DataRepo)]
    generate(all_repos , dataname; kwargs...)
end


function generate(repos, dataname, shortname = nothing, show_failures=false)
	retrieved_metadatas = MetaData[]
	failures = Tuple{DataRepo, Exception}[]
	
	# Get all the metadata we can
	@sync for repo in repos
		@async try
			metadata = find_metadata(repo, dataname, shortname)
			push!(retrieved_metadatas, metadata)
		catch err
			push!(failures, (repo, err))
		end
	end
	
	# Displace errors if required
	if length(retrieved_metadatas) == 0 || show_failures
		for (repo, err) in failures
			println(repo, " failed due to")
			println(err)
			println()
		end
	end
	
	# merge them
	aggregate(retrieved_metadatas)
end
```

I haven't tested this code.
But hopefully you can get the idea.


-------------------
Please consider and incorperate something like this.
I believe this is the best way  to handle calling multiple generators
and allowing some to fail.

yes, it involved catching all exceptions.
But that is ok at the top-level.
It is an UX thing, since we are capturing them,
and only displaying them if they matter (i.e. if all generators failed.)

src/DataCite.jl
oxinabox: Don't forget to delete this

src/DataDepsGenerators.jl
```diff
@@ -30,26 +30,32 @@ end
...
+            modified_date(repo, mainpage),
+            paper_cite(repo, mainpage),
+            dataset_cite(repo, mainpage),
+            get_urls(repo, mainpage),
+            get_checksums(repo, mainpage)
+        )
+    # catch err
+    #     # if isa(err, GeneratorError)
+    #     # end
+    # end
```
oxinabox: Don't forget to delete these commented out lines

src/DataDepsGenerators.jl
```diff
@@ -125,14 +174,57 @@ function format_meta(data::Any, label="")
...
+#     metadatas = []
+#     push!(generators, repo)
+#     for ii in generators
+#         meta = find_metadata(ii, dataname, shortname)
+#         if !isa(meta, Void)
+#             push!(metadatas, meta)
+#         end
+#     end
+#     aggregate(metadatas)
+# end
```
oxinabox: Do not forget to delete commented out code

src/DataDepsGenerators.jl
```diff
@@ -87,6 +93,49 @@ include("DataCite.jl")
...
+
+combine(::Missing, ::Missing, z) = missing
+combine(::Missing, x, z) = x
+combine(x, ::Missing, z) = x
+combine(x::Vector,y::Vector, z) = length(x) > length(y) ? x : y
+combine(x::String,y::String,s::Symbol) = combine(x,y,Val{s}())
+combine(x::String,y::String, ::Val{:license}) = length(x) < length(y) ? x : y
+combine(x::String,y::String, z) = length(x) > length(y) ? x : y
+combine(x::Union{DateTime, Date},y::String, z) = x
+combine(x::String,y::Union{DateTime, Date}, z) = y
```
oxinabox: If you do not use a parameter, e.g. `z` don't give it a name.

you can write `combine(::Missing, ::Missing, ::Any)`
which is clearer.

src/DataDepsGenerators.jl
```diff
@@ -87,6 +93,49 @@ include("DataCite.jl")
...
+        ret = combine(ret, value, sym)
+    end
+    ret
+end
+
+combine(::Missing, ::Missing, z) = missing
+combine(::Missing, x, z) = x
+combine(x, ::Missing, z) = x
+combine(x::Vector,y::Vector, z) = length(x) > length(y) ? x : y
+combine(x::String,y::String,s::Symbol) = combine(x,y,Val{s}())
```
oxinabox: Move this one to the top away,
with a line space below it.
And a comment saying something like "redispatch based on value".

Also I think the better signature is:

```
combine(x::T1, y::T2, s::Symbol) where {T1, T2} = combine(x, y, Val{s}())
```

Also the spacing in general around `,` is not great in this section.
basically after a `,` there should always be a space.

src/DataDepsGenerators.jl
```diff
@@ -125,14 +174,57 @@ function format_meta(data::Any, label="")
...
+	@sync for repo in repos
+		@async try
+			metadata = find_metadata(repo, dataname, shortname)
+			push!(retrieved_metadatas, metadata)
+		catch err
+			push!(failures, (repo, err))
+		end
+	end
+	
+	# Displace errors if required
```
oxinabox: `# Display errors if required`

src/DataDepsGenerators.jl
```diff
@@ -125,14 +174,57 @@ function format_meta(data::Any, label="")
...
+		for (repo, err) in failures
+			println(repo, " failed due to")
+			println(err)
+			println()
+		end
+	end
+	
+	# merge them
+	aggregate(retrieved_metadatas)
 end
```
oxinabox: Indentation in this section has been done with tabs.
Everything else is indented with spaces.

I suspect this is because the code I gave you in the comment was that way.
There should be a `retab`  or `expandtabs` or similar command in your editor that will fix this.

src/DataDepsGenerators.jl
```diff
@@ -190,4 +282,10 @@ function mainpage_url(repo::DataRepo, dataname)
     getpage(url), url
 end
 
+struct GeneratorError <: Exception
+    repo::DataDepsGenerators.DataRepo
+end
```
oxinabox:
```
struct GeneratorError{T<:DataRepo} <: Exception
    repo::T
end
```

Don't put abstract types on fields, if it can be avoided, without making the code too ugly.
https://docs.julialang.org/en/stable/manual/performance-tips/#Avoid-fields-with-abstract-type-1

it isn't a major performance issue here, but it is good practice.

src/DataOneV2/DataOneV2.jl
```diff
@@ -4,30 +4,36 @@ export KNB, TERN
...
 
-license(repo::DataOneV2, mainpage) = license_(repo, mainpage)
+license(repo::DataOneV2, mainpage) = try license_(repo, mainpage) catch missing end
 
-published_date(repo::DataOneV2, mainpage) = pub_date(repo, mainpage)
-
-get_urls(repo::DataOneV2, page) = []
+published_date(repo::DataOneV2, mainpage) = try pub_date(repo, mainpage) catch missing end
 
 function license_(repo::DataOneV2, mainpage)
```
oxinabox: rename this to 
`function license(repo::DataOneV2, mainpage)`

and delete the definition on line 11.

The others also need the same done, but that can come later.
This one is particularly bothersome,
as it is double wrapped in `try`

src/JSONLD/JSONLD_Web.jl
```diff
@@ -1,14 +1,18 @@
...
+        jsonld_blocks = matchall(pattern, page.root)
+        if length(jsonld_blocks)==0
+            error("No JSON-LD Linked Data Found")
+        end
+        @assert length(jsonld_blocks)==1
+        script_block = text_only(first(jsonld_blocks))
+        json = JSON.parse(script_block)
+        return json, dataname
+    catch
+        throw(GeneratorError(repo))
```
oxinabox: Delete the `try`/`catch` here.
The error that is thrown is more informative anyway.
We want to show the error to the user at the top level,
so that if they were expecting this to work, they know why it didn't.
Thus we do not want to hide the original exception message.


You could add an optional message field (default to `""`) to `GeneratorError` and,
change line 9, to be 
`throw(GeneratorError, repo, "No JSON-LD Linked Data Found")`

src/DataDepsGenerators.jl
oxinabox: Maybe give this a `message::String` field,
with default value `""`,
(so add the constructor `GeneratorError(repo)=GeneratorError(repo, "")`

And add that to the message to be shown during `showerror`.

src/DataDepsGenerators.jl
```diff
@@ -190,4 +282,10 @@ function mainpage_url(repo::DataRepo, dataname)
...
+
+Base.showerror(io::IO, e::GeneratorError) = print(io, e.repo, " generator did not work")
```
oxinabox: I feel better that `did not work`,
would be `was not suitable`.
Since this is the error thrown if it is detected that it could not be used, e.g via not being a DOI.
Obviously every error means that it did not work

Or if you add a `message` field, then it would be `was not suitable. $(e.message)`

test/JSONLD/JSONLD.jl
oxinabox: Hopefully now this can be uncommented out?
If not needs to be fixed before this can be merged

test/async.jl
```diff
@@ -0,0 +1,12 @@
+using DataDepsGenerators
+using Base.Test
+
+using ReferenceTests
+
+@testset "DataCite test" begin
```
oxinabox: I think maybe this is not the right test set title.

Also the file name is wrong.

Async is not what you are testing here -- combining is.
It uses async, but not in a way you can test for.

test/references/Async Ecology.txt
```diff
@@ -0,0 +1,18 @@
+register(DataDep(
```
oxinabox, line 1: YESSS! 
AWESOME!

I am looking at this compared to 
https://github.com/oxinabox/DataDepsGenerators.jl/blob/master/test/references/DataCite%20Ecology.txt

and

https://github.com/oxinabox/DataDepsGenerators.jl/blob/master/test/references/DataDryad%20Ecology.txt

and it has the best parts of  each!

test/references/DataCite Ceramic.txt
```diff
@@ -10,6 +10,5 @@ register(DataDep(
...
 	
 	Please cite this dataset: Arora, R. (2018). Study Of Surface Roughness With The Variation In Applied Load Of Rutile Ceramic Reinforced Aluminium Composite. https://doi.org/10.5281/zenodo.1147572
-	""",
-	String["PUT DOWNLOAD URL HERE"],
+	""",,
```
oxinabox: hmmm, 
there should be a `missing`  on this line where the URL should be.
I think that would be clearest.

oxinabox (commented):
Woooot!
We are on the home stretch now.

Some cleanup left,
but the core of the logic is basically all there

test/async.jl
```diff
@@ -0,0 +1,12 @@
...
+    @test_reference "references/Async Ecology.txt" remove_cite_version(generate("10.5061/dryad.74699"))
+    
+    @test_reference "references/Async Plasticity.txt" remove_cite_version(generate("https://datadryad.org/mn/object/http://dx.doi.org/10.5061/dryad.f9s4424"))
+
+    @test_reference "references/Async Gene Diversity.txt" remove_cite_version(generate("10.5281/zenodo.1194927"))
```
oxinabox: Add a test that uses FigShare.

src/DataDepsGenerators.jl
oxinabox: Solved now

src/DataDepsGenerators.jl
```diff
@@ -126,15 +119,17 @@ function combine_all(values::Vector, sym::Symbol)
...
-combine(::Missing, x, z) = x
-combine(x, ::Missing, z) = x
-combine(x::Vector,y::Vector, z) = length(x) > length(y) ? x : y
-combine(x::String,y::String,s::Symbol) = combine(x,y,Val{s}())
-combine(x::String,y::String, ::Val{:license}) = length(x) < length(y) ? x : y
-combine(x::String,y::String, z) = length(x) > length(y) ? x : y
-combine(x::Union{DateTime, Date},y::String, z) = x
-combine(x::String,y::Union{DateTime, Date}, z) = y
+#Redispatch based on Value
+combine(x::String, y::String, s::Symbol) = combine(x,y,Val{s}())
```
oxinabox: Did 
```
combine(x::T1, y::T2, s::Symbol) where {T1,T2} = combine(x,y,Val{s}())
```
not work out?

src/DataDepsGenerators.jl
```diff
@@ -157,7 +152,7 @@ function body(meta)
     netString *= "\n\t\"\"\","
     netString = netString |> strip
     ## End of the message
-    netString *= format_meta(meta.dataurls)
+    netString *= format_meta(ismissing(meta.dataurls)?"missing":meta.dataurls)
```
oxinabox: Spacing is  needed around `?` and `:`

src/DataDepsGenerators.jl
oxinabox: sorted now

src/DataDepsGenerators.jl
oxinabox: sorted now

src/DataDepsGenerators.jl
oxinabox: sorted now

src/JSONLD/JSONLD_Web.jl
```diff
@@ -1,18 +1,14 @@
...
-        script_block = text_only(first(jsonld_blocks))
-        json = JSON.parse(script_block)
-        return json, dataname
-    catch
-        throw(GeneratorError(repo))
+    page=getpage(dataname)
+    pattern = sel"script[type=\"application/ld+json\"]"
+    jsonld_blocks = matchall(pattern, page.root)
+    if length(jsonld_blocks)==0
+        throw(GeneratorError, repo, "No JSON-LD Linked Data Found")
```
oxinabox: No that is not right, throw doesn't work like that. Does it?

```
 throw(GeneratorError(repo, "No JSON-LD Linked Data Found"))
```

test/combine.jl
```diff
@@ -0,0 +1,12 @@
...
+
+using ReferenceTests
+
+@testset "Combining test" begin
+    @test_reference "references/Combining Ecology.txt" remove_cite_version(generate("10.5061/dryad.74699"))
+    
+    @test_reference "references/Combining Plasticity.txt" remove_cite_version(generate("https://datadryad.org/mn/object/http://dx.doi.org/10.5061/dryad.f9s4424"))
+
+    @test_reference "references/Combining Gene Diversity.txt" remove_cite_version(generate("10.5281/zenodo.1194927"))
+end
```
oxinabox: Still need to add a Figshare test here.
Maybe a CKAN one too just to make sure it doesn't bork.

test/JSONLD/JSONLD.jl
oxinabox: This still needs to be uncommented out

test/references/DataCite Ceramic.txt
oxinabox: Should be sorted now, but you need to add the newly generated files

oxinabox (commented):
Cool cool
getting close to done.

oxinabox:
Right the version I just pushed works.
I disabled the Async.
That is not the important part of this PR anyway.
It may be an actual julia bug.
If so, solving it is better to do after upgrading to julia 0.7.
So lets leave it for now

Every thing else I said still need to be done still need to be done.


Also I know what is wrong with the JSON-LD for figshare and I've risen an issue.
it can stay commented out for now.

src/DataDepsGenerators.jl
SebastinSanty: Nope, it didn't, cause it was ambigous with `combine(x::String, y::String, ::Any)` and others.

oxinabox: hmm might want to ask on #helpdesk or discourse about avoiding that.

oxinabox (approved):
Oh cool, cool.
I just checked over this, 
I think it is good to go.

You can merge it if you agree.
I'll be presenting on it tomorrow.

SebastinSanty:
Oh great! I was a bit confused too. Still I was reading up on https://docs.julialang.org/en/stable/manual/methods/index.html#man-methods-orthogonalize-1.

oxinabox:
An excellent thing to be reading.

PR #49, opened 2018-08-21
Documentation in README

SebastinSanty:


README.md
```diff
@@ -101,21 +101,30 @@ Then just use anywhere in your code (later in the REPL session for example)  `da
...
 
+## Wildcard Use
 
+The above examples show the usage of `generate()` with a DataRepo. There are instances when the user isn’t sure of the data generators to use or if the user wants to reap maximum benefits from all the generators. Hence in order to facilitate that, you can use `generate()` without providing the DataRepo as an argument:
 
+```julia
+    generate("10.5061/dryad.74699")
+```
+
+This will scrounge in all the available supported DataRepos asynchronously to get the best of all the data according to rules defined.
```
oxinabox: Cut the `asynchronously`  since that is not true.
I would also say this is the normal or default usage.
And point out that it may give better metadata than you expect,
as it may find sources you were unaware of.

README.md
```diff
@@ -101,21 +101,30 @@ Then just use anywhere in your code (later in the REPL session for example)  `da
 (Which indeed what that string macro expands into -- even if it has to download the data first).
 
 
+## Wildcard Use
 
+The above examples show the usage of `generate()` with a DataRepo. There are instances when the user isn’t sure of the data generators to use or if the user wants to reap maximum benefits from all the generators. Hence in order to facilitate that, you can use `generate()` without providing the DataRepo as an argument:
 
+```julia
+    generate("10.5061/dryad.74699")
```
oxinabox: Don't use this as an example.
Use something with an URL.
A URL is the most common identifier.

oxinabox (commented):
Looks good.
Can you add a link to the reference file in the tests for each example

oxinabox:
How is this going?
(How are your other things going also BTW?)

SebastinSanty:
My other things are going fine! I got a bit busy with some work I had to finish.

SebastinSanty:
Is the README good?

oxinabox:
Looks good.
To be perfect would be to have links to example outputs from our reference tests.
But that can come another time.
It is much better than the current docs

PR #51, opened 2018-08-28
Reenable Figshare JSON-LD test

oxinabox:
FigShare support inform me they have fixed the bug in their JSON-LD generator.
And it does indeed seem to be fixed


oxinabox:
Seems like the new file is worst than the old.  (Check the diff v:-))
Was passing for me before.
Hmmm, what has changed

PR #53, opened 2018-09-04
Quieten Integration tests

oxinabox:
@andreasnoack  informs me that this is dumping so much console output that it is crashing the JuliaCIBot server.

I believe this is due to the fact that `Base.download` updates it's status to the console nearly constantly.

This should fix that


oxinabox:
this seems to be failing as there has been a regression in the outputs.
Perhaps due to an service changing APIs

oxinabox:
------------------------------------------

So,
https://figshare.com/articles/Supplemental_Material_for_Overcoming_challenges_in_variant_calling_exploring_sequence_diversity_in_candidate_genes_for_plant_development_in_perennial_ryegrass_Lolium_perenne_/5969383

has been deleted from figshare,
so we need to replace that test.

-------------------------------------------

"DataCite Ceramic.txt"
should be fine
https://api.datacite.org/works/10.5281/zenodo.1147572
works fine

------------------------------------------------


Test for "CKAN Consumer Complaint.txt" failed.
should be fine

https://catalog.data.gov/api/3/action/package_show?id=consumer-complaint-database
seems fine.
(if slow)

I have been been firewalled out of Data.gov before though, (with A.V)
So that could have happened again.

------------------------------------------------

PR #58, opened 2018-09-15
Upgrate to 0.7/1.0 (attempt 3)

oxinabox:
Trying to get femtocleaner.
This replaces #55

PR #55, opened 2018-09-14
Bump to 0.7 /1.0

oxinabox:
replaces #48

PR #60, opened 2018-09-15
Full 1.0 compat

oxinabox:
Finish up from #59  #58  #55 etc


oxinabox:
This should be alll good now for 1.0

oxinabox:
ok,  now datadryad is down for everyone

oxinabox:
This is all done now.
(Final CI is still running but it got past the only change of relevance from the last time it passed)

Issue #67, opened 2018-10-22
Some README examples failing


ghost:
Some appear to be HTTP issues and others possible syntax changes in Gumbo.

```> julia -v
julia version 1.0.1```

```

# DataOneV1 #

```
julia> generate(DataOneV1(), "https://datadryad.org/resource/doi:10.5061/dryad.74699", "Wild Crop Genomics")
ERROR: BoundsError: attempt to access 0-element Array{Gumbo.HTMLNode,1} at index [1]
Stacktrace:
 [1] getindex at ./array.jl:731 [inlined]
 [2] first at ./abstractarray.jl:270 [inlined]
 [3] data_fullname(::DataOneV1, ::Gumbo.HTMLDocument) at /.julia/packages/DataDepsGenerators/CbcdV/src/APIs/DataOneV1.jl:52
 [4] find_metadata(::DataOneV1, ::String, ::String) at /.julia/packages/DataDepsGenerators/CbcdV/src/generate.jl:73
 [5] #generate#7 at /.julia/packages/DataDepsGenerators/CbcdV/src/generate.jl:25 [inlined]
 [6] generate(::DataOneV1, ::String, ::String) at /.julia/packages/DataDepsGenerators/CbcdV/src/generate.jl:25
 [7] top-level scope at none:0
```

# DataOneV2 #

```
julia> generate(KnowledgeNetworkforBiocomplexity(), "https://knb.ecoinformatics.org/knb/d1/mn/v2/object/doi:10.5063/F1T43R7N")
ERROR: IOError(EOFError() during request(https://knb.ecoinformatics.org/knb/d1/mn/v2/object/doi:10.5063/F1T43R7N))

Stacktrace:
 [1] readuntil(::HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}, ::typeof(HTTP.Parsers.find_end_of_header)) at /.julia/packages/HTTP/YjRCz/src/IOExtras.jl:168
 [2] readheaders(::HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}, ::HTTP.Messages.Response) at /.julia/packages/HTTP/YjRCz/src/Messages.jl:469
 [3] startread(::HTTP.Streams.Stream{HTTP.Messages.Response,HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}}) at /.julia/packages/HTTP/YjRCz/src/Streams.jl:149
[...]

PR #71, opened 2018-10-26
Delete DataOneV1

oxinabox:
Re https://github.com/oxinabox/DataDepsGenerators.jl/issues/67#issuecomment-432897192


oxinabox:
the combining failure is because now there is one less generator and the results are different.
Semantically they are the same, but since the different providers do things like `http` vs `https` for the website that does change the reference.
So that needs to be regenerated.

oxinabox:
while fixing this I noticed that our DataDryad generator wasn't working for DOIs, which I fixed,
and added tests.

Also I think this will fail, as I think the DataCite API is down
cf: https://twitter.com/oxinabox_frames/status/1056115555499823105
Update: yes indeed it is down.

oxinabox:
Tests passing.
@SebastinSanty  do you want to review it?

SebastinSanty:
Sure, I'll check

SebastinSanty (approved):
LGTM

